Issue #239, opened 2026-04-11
quant-lab → OpenQuant replay: Jan-Apr 2026 produces +$3,399 (74% win), March alone -$499


kumargu:
### Division of labor

> Lab does a good job finding raw eligible pairs. OpenQuant does a good job filtering and trading it neatly.

- **Quant-lab** = candidate source. Scans 124,251 S&P 500 pairs with fixed-formation z-score. Finds 3,000-8,000 profitable pairs per 3-month window. Ranks by realized total_bps. Top 100 is the productive candidate pool.
- **OpenQuant Rust picker** = safety filter. Structural scoring (ADF p-value, half-life, R², β stability) filters out lucky-trade noise. Proven: taking lab's top-40 blindly → **−\$2,809**; letting Rust pick its top-40 → **+\$3,399** on the same candidates.
- **OpenQuant engine** = trader. 1-min bars, rolling z-score, stop-loss, max-hold, weekly pair regeneration. Handles intraday noise daily-bar backtests can't see.

### Baseline results — 3 disjoint OOS windows

| window | formation | trades | win% | P&L | \$/trade |
|---|---|---|---|---|---|
| May-Jul 2025 | Jan-Apr 2025 (80d) | 56 | 76.8% | +\$4,364 | +\$78 |
| Sep-Nov 2025 | Jan-Aug 2025 (168d) | 30 | 66.7% | +\$2,486 | +\$83 |
| Jan-Apr 2026 | Feb 2025-Dec 2025 (216d) | 38 | 73.7% | +\$3,399 | +\$89 |
| **Aggregate (9 months)** | | **124** | **73.4%** | **+\$10,249** | **+\$83** |

- Binomial p-value for 91/124 wins vs 50% null ≈ **2e-8**
- Different top pairs in each window (FSLR/VST, EXPE/INCY, APA/KEYS) — pair rotation is structural
- Exit mix stable: ~65% clean reversion, ~15% max-hold, ~15-20% stop-loss

### Key findings

**1. Rust picker's structural score is essential — not broken**

The earlier hypothesis that "the picker is scoring the wrong objective" was wrong. The preserve_input_order experiment proved the opposite:

| config | trades | win% | P&L |
|---|---|---|---|
| Rust sort (default) | 38 | 73.7% | **+\$3,399** |
| preserve_input_order=true | 29 | 51.7% | **−\$2,809** |

[...]

PR #240, opened 2026-04-12
Remove pair-picker discovery code — lab owns candidate selection

kumargu:
## Summary

- Pair-picker is now **validation-only** — quant-lab discovers and ranks candidates, Rust validates structural quality (ADF, R², HL, β stability)
- Removed 2,031 lines of dead discovery code (Thompson sampling, lockfile, standalone binary, relationship graph, regime gate)
- Added `[pair_picker]` TOML config (`top_k`, `preserve_input_order`) and `ALPACA_DATA_URL` env override for offline replay
- New `scripts/mock_alpaca.py` serves bars from parquets for credential-free offline replay

## Backtest comparison

Baseline replay (Jan-Apr 2026, top-100 lab candidates, `--pipeline force`, `top_k=40`):

| metric | before cleanup | after cleanup |
|---|---|---|
| trades | 38 | 38 |
| wins / losses | 28 / 10 | 28 / 10 |
| win rate | 73.7% | 73.7% |
| total P&L | +$3,399 | +$3,399 |

**Identical.** Also verified after each intermediate deletion (graph.rs, regime.rs).

Full 3-window OOS validation (see #239):

| window | trades | win% | P&L |
|---|---|---|---|
| May-Jul 2025 | 56 | 76.8% | +$4,364 |
| Sep-Nov 2025 | 30 | 66.7% | +$2,486 |
| Jan-Apr 2026 | 38 | 73.7% | +$3,399 |
| **Aggregate** | **124** | **73.4%** | **+$10,249** |

## What was removed

| file | lines | purpose |
|---|---|---|
| `thompson.rs` | 728 | Thompson sampling for pair ranking |
| `regime.rs` | 480 | Regime robustness gate (was returning -1.0 in production; deferred to lab) |
| `main.rs` | 318 | Standalone CLI binary |
| `graph.rs` | 306 | Relationship graph → candidate generation |
| `lockfile.rs` | 161 | Date-based run-once lock |
| `[[bin]]` + `test-support` | — | Build config for deleted binary |
| Graph/default fallbacks in service | ~40 | 3 candidate loading paths → 1 |

## What remains (validation pipeline)

| file | purpose |
|---|---|
| `pipeline.rs` | Validation orchestrator (ADF, R², HL, β stability, ETF, score, sort, truncate) |
| `scorer.rs` | Structural quality scoring formula |
| `stats/` | ADF, OLS regression, half-life estimation |
| `etf_filter.rs` | ETF-component pair exclusion |
| `types.rs` | PairCandidate, ActivePair, PairCandidatesFile schemas |
| `test_utils.rs` | `#[cfg(test)]` synthetic data generators |

## Test plan

- [x] `cargo test -p pair-picker` — 115 tests pass (102 unit + 7 integration + 6 doc-tests)
- [x] `cargo test -p openquant-runner` — 5 tests pass
- [x] `cargo build -p openquant-runner --release` — clean
- [x] Replay baseline verified identical before and after each deletion
- [ ] CI (cargo fmt, clippy, full workspace test)


engine/crates/runner/src/main.rs
```diff
@@ -323,13 +323,13 @@ async fn run(
...
     // ── Resolve pipeline config ──
-    let pipeline_cfg = match pipeline_profile.as_str() {
+    let mut pipeline_cfg = match pipeline_profile.as_str() {
         "metals" => {
             info!("using METALS pipeline thresholds");
             PipelineConfig::metals()
         }
-        "force" => {
-            info!("using FORCE pipeline — all pairs pass validation");
+        "force" | "lab" => {
```
kumargu: remove force ?

kumargu: maybe call replay ? which uses mock server

engine/crates/runner/src/main.rs
```diff
@@ -338,17 +338,51 @@ async fn run(
...
     let mut ptc = cfg_file.pairs_trading.clone();
     ptc.tz_offset_hours = cfg_file.data.timezone_offset_hours;
 
     let history_path = trading_dir.join("pair_trading_history.json");
 
-    let active_pairs_path = trading_dir.join("active_pairs.json");
+    // Active pairs file: monthly_pairs_YYYYMM.json (lab provides one per month).
+    // Replay derives YYYYMM from the start date; live/paper uses the current month.
+    // Falls back to active_pairs.json for backward compatibility.
+    let active_pairs_path = {
```
kumargu: write tests

engine/Cargo.toml
```diff
@@ -1,3 +1,3 @@
 [workspace]
-members = ["crates/core", "crates/pybridge", "crates/journal", "crates/metrics", "crates/pair-picker", "crates/runner"]
+members = ["crates/core", "crates/pybridge", "crates/journal", "crates/metrics", "crates/pair-picker", "crates/runner", "crates/oracle"]
```
kumargu: do we need oracle ? we do all the lab work in a different repo now. we can revove all oracle refresnces ?

scripts/mock_alpaca.py
```diff
@@ -0,0 +1,220 @@
+"""
+Mock Alpaca market-data server — serves bars from the quant-data parquets.
```
kumargu, line 2: weite in rust and keep it close to alcapa.rs . use enums to select modes. write tests.

scripts/mock_alpaca.py
```diff
@@ -0,0 +1,220 @@
...
+
+Bypasses the real Alpaca API during offline replay. The runner's AlpacaClient
+(engine/crates/runner/src/alpaca.rs) can be pointed here via the
```
kumargu, line 5: also wondering why is mock so complex does it alogn exaclyy with how lab uses its data from openquant data ?

autoresearch/ORACLE_SPEC.md
```diff
@@ -0,0 +1,213 @@
+# Oracle Spec (Mode A) — locked v1
+
```
kumargu: this might be stale now. the oracle is now the lab.

autoresearch/LEARNINGS.md
```diff
@@ -0,0 +1,13 @@
+
+## Bug: Duplicate symbol exposure across pairs
```
kumargu: delete

autoresearch/PAIR_PICKER_CONSUMER_CONTRACT.md
```diff
@@ -0,0 +1,154 @@
+# Pair-Picker Consumer Contract Inventory
+
+Everything the rest of the workspace imports from the `pair-picker` crate.
+This is the surface that must be preserved (or deliberately broken and
+rewritten) when we delete the old pipeline.
+
+## External consumers (must be handled)
+
+### `engine/crates/runner/src/pair_picker_service.rs` (148 lines)
+The runtime integration layer. This file wraps the pair-picker for the
```
kumargu: delete this file

autoresearch/PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+---
+
+## Autoresearch Mode
+- Run autonomously, don't stop
+- Commit after every major finding
+- Pause at milestones to check plan progress
+- No code changes in Phase 1 — only analysis scripts and findings
+- Log everything in pattern_research.md
+
+## Phase 1: CHEAT (read-only, learn from bar cache)
```
kumargu: no longer needed, cheatin happesn in lab. try to now see what lab does and what this repo does now. it gets pairs ad run experiment with varpus congogs and code chnages with various algothms over pairs

kumargu:
## Review — PR #240: cleanup/pair-picker-validation-only

### CI: FAILING — do not merge

`cargo fmt --check` fails in four files. Fix required before merge.

```
engine/crates/runner/src/alpaca.rs
  line 97:  RTH_END_MINUTES alignment spacing
  line 175: fetch_minute_bars_raw call needs wrapping

engine/crates/runner/src/main.rs
  line 360: RunMode::Stream arm should be single-line

engine/crates/runner/src/pair_picker_service.rs
  line 34:  read_to_string chain needs wrapping

engine/crates/oracle/src/main.rs
  line 29:  import ordering (ArrowWriter / ParquetRecordBatchReaderBuilder)
  line 175: NaiveDate::parse_from_str calls need wrapping
```

Run `cargo fmt --all` locally and push a fixup commit.

---

### Issue responses (inline comments from @kumargu)

**1. Remove `force` alias — keep only `lab` (main.rs line 331)**

Agree. The `"force" | "lab"` arm in the `pipeline_profile` match is a confusing artefact. The name `force` implies 'override everything', which is not how the `lab` profile should be described to operators running live commands. However: `Engine::Metals` still returns `"force"` from `pipeline()` (line 85). Before removing the alias, either rename `PipelineConfig::force()` → `PipelineConfig::lab()`, or give Metals its own distinct profile name (`"metals_force"` is fine). Right now removing `"force"` from the match arm would silently route Metals runs to `error!(unknown pipeline)` and exit. Fix the Metals arm first, then remove the alias.

**2. Write tests for the pair-picker service (main.rs line 357)**

Agreed and overdue. The service is the integration seam between the runner and pair-picker crate — it is the most likely place for a contract breakage to go undetected. Required test cases:
- `generate_pairs_with_config` with `candidates_path = None` returns `Err` (the required-flag path)
- Round-trip: synthetic candidates JSON → `generate_pairs_with_config` → `write_active_pairs` → file readable and non-empty
- `to_pair_configs` with `half_life_days = 0.0` and `half_life_days = f64::INFINITY` does not panic and sets `kappa = 0.0`

These three cover the edge paths that silent data corruption would hit first.

**3. Remove oracle crate (Cargo.toml line 2)**

The oracle crate is a standalone binary that implements the brute-force ground-truth classifier. It is not dead code — it is the mechanism for generating the `verdicts.parquet` files that autoresearch uses to evaluate pickers. It depends on `openquant_core` directly so there is no drift between oracle strategy and production strategy. I would not remove it. That said, if quant-lab now owns ground-truth evaluation, the oracle may become redundant. That decision belongs in a separate issue with a clear replacement plan, not in this cleanup PR. Leave it in for now.

**4. Rewrite mock_alpaca.py in Rust (scripts/mock_alpaca.py line 2)**

Directionally correct, but this should be a follow-up issue, not a blocker for this PR. The current Python server is simple, well-understood, and serves one purpose: translate parquet reads into Alpaca HTTP responses. Rewriting in Rust would add value (type safety on the Bar struct, aligned RTH filter, `#[cfg(test)]` integration tests), but it is a clean engineering task that does not need to block the cleanup. File a follow-up issue.

**5. mock_alpaca.py complexity question (scripts/mock_alpaca.py line 5)**

The mock handles both `1Day` and `1Min` timeframes. The daily aggregation in the mock mirrors `aggregate_to_daily` in `alpaca.rs`: RTH filter 13:30–20:00 UTC, last tick per session. This is intentional and correct — if the mock returned raw 1-min bars for `1Day` requests, the pair-picker would receive unaggregated data and its statistics would diverge from quant-lab. The complexity is justified. One concrete issue: the mock's `1Day` response timestamps bars at midnight UTC (`00:00Z`), but `fetch_daily_bars` in `alpaca.rs` snaps warmup timestamps to `20:00 UTC`. For replay, pair-picker calls `fetch_daily_bars_range` (which discards timestamps entirely, keeping only closes), so this is harmless today. It would become a bug if anything downstream ever inspects the timestamp field of daily bars from the mock. Worth a comment in the mock explaining the discrepancy.

**6. Delete ORACLE_SPEC.md, LEARNINGS.md, PAIR_PICKER_CONSUMER_CONTRACT.md, PLAN.md**

- **LEARNINGS.md**: Delete. Contains one bug note (duplicate symbol exposure across pairs) that should be a GitHub issue, not a stray markdown file with no owner.
- **PAIR_PICKER_CONSUMER_CONTRACT.md**: Delete. Describes the pre-PR crate API. Now stale — the new consumer contract is `validate_candidates_with_config` and `to_pair_configs`, both in this PR.
- **PLAN.md**: Delete. Describes the Phase 1/2/3 experiment plan that has since been superseded by the lab→picker→replay architecture documented in AUTORESEARCH_LOOP.md.
- **ORACLE_SPEC.md**: Do not delete yet. The oracle crate implements this spec and the spec contains non-obvious locked decisions (classification thresholds, 5 bps cost model, INACTIVE cutoff). If the oracle crate stays, the spec stays. If the oracle crate is removed in a follow-up, delete both together.

---

### Independent findings (not from inline comments)

**Dead code: `run_pipeline`, `run_pipeline_from_candidates`, `run_pipeline_from_candidates_with_config`**

These three public functions in `pipeline.rs` (lines 574–690) write to an `output_path` and were the binary's entry points. After removing the `main.rs` binary, no external caller uses them. The runner calls `validate_candidates_with_config` only. Internal callers are test fixtures at lines 891 and 967 (both inside `#[cfg(test)]`). These functions are effectively dead public API. They should be removed or demoted to `#[cfg(test)]`. Leaving them in the public API implies callers exist and that the 'write to file' path is the canonical way to call the pipeline — it isn't.

**Stale comment at Engine::Snp500 candidates_path (main.rs line 77)**

```rust
Engine::Snp500 => None, // uses stock_relationships.json / pair_candidates.json
```

`stock_relationships.json` and `pair_candidates.json` no longer exist — they belonged to the deleted graph.rs / lockfile.rs path. The correct comment is: `None — candidates must be provided via `--candidates` flag`.

**NaN guard missing in aggregate_to_daily**

`alpaca.rs` `aggregate_to_daily` stores `bar.c` without checking `bar.c.is_finite() && bar.c > 0.0`. The mock server explicitly sets missing values to `0.0` (`float(c[i] or 0.0)`), meaning a missing close in the parquet produces a 0.0 close that flows into ADF/OLS as a legitimate observation. The pipeline's NaN guard at lines 251/258 of `pipeline.rs` catches non-positive prices, so the damage is contained — the pair gets rejected. But the correct place to reject is at the data boundary, not four layers deep. Add a guard in `aggregate_to_daily`:

```rust
if !bar.c.is_finite() || bar.c <= 0.0 {
    continue; // skip malformed bars rather than propagating 0.0
}
```

**Monthly file naming — live mode uses current UTC month, not local**

`RunMode::Stream(_) => chrono::Utc::now().format("%Y%m").to_string()`

If the system is running in a non-UTC timezone (e.g., US/Eastern, UTC-4), a run started at 11:30 PM local time on March 31 would derive `202404` and look for `monthly_pairs_202404.json` when the correct file is `monthly_pairs_202603.json`. This is a real risk in live trading. Either document that the system must run in UTC, or derive YYYYMM from the local date explicitly.

**`test_utils` is `#[cfg(test)]` only — integration tests duplicate generators**

`lib.rs` marks `test_utils` as `#[cfg(test)]` only. The integration test file `tests/integration_pipeline.rs` cannot import it and duplicates `cointegrated_pair` and `independent_walks` verbatim. The CLAUDE.md standard is: shared generators in a `#[cfg(test)]` module, reused across all test files. This can be fixed by adding a `test-support` feature flag that exposes `test_utils` publicly when enabled (the deleted Cargo.toml build config mentioned this feature). Alternatively, since the integration test generators are only ~50 lines, a comment acknowledging the deliberate duplication is acceptable. But the duplication should be called out, not silently left.

---

### Summary

**Required before merge:**
1. Fix `cargo fmt` failures (four files — run `cargo fmt --all` and push)
2. Fix the `Engine::Metals => "force"` routing before removing the `"force"` alias (or keep the alias — but if keeping it, remove the owner's inline comment requesting its removal)
3. Remove `run_pipeline` / `run_pipeline_from_candidates` / `run_pipeline_from_candidates_with_config` from the public API or add `#[allow(dead_code)]` with a comment explaining why they're kept
4. Fix the stale `stock_relationships.json` comment at main.rs line 77
5. Delete LEARNINGS.md, PAIR_PICKER_CONSUMER_CONTRACT.md, PLAN.md (per owner's own inline comments — they should not be in the PR if the owner already decided to delete them)

**Follow-up issues to file after merge:**
- Rust mock server replacement for mock_alpaca.py
- Tests for pair_picker_service
- UTC-only requirement for live mode (or local-date fix)
- NaN guard in aggregate_to_daily

The backtest numbers are consistent (73.7% win rate, +$3,399 identical before/after), the architectural direction is correct, and the 115 tests passing validate the remaining pipeline. This PR is on the right track but not yet mergeable.

engine/crates/runner/src/main.rs
```diff
@@ -338,17 +342,49 @@ async fn run(
...
+            RunMode::Stream(_) => chrono::Utc::now().format("%Y%m").to_string(),
+        };
+        let monthly = trading_dir.join(format!("monthly_pairs_{yyyymm}.json"));
+        let legacy = trading_dir.join("active_pairs.json");
+        if monthly.exists() {
+            info!(path = %monthly.display(), "using monthly pairs file");
+            monthly
+        } else if legacy.exists() {
+            info!(path = %legacy.display(), "monthly_pairs not found, falling back to active_pairs.json");
+            legacy
```
kumargu: no legacy. remove active_pairs and its references? remove any reference of file which has pair data

kumargu:
## Code Review — cleanup/pair-picker-validation-only

CI is red. Two blockers plus cosmetic issues.

---

### BLOCKER 1: Clippy failures (5 errors in `engine/crates/core/src/pairs/mod.rs`)

**`doc_lazy_continuation` — lines 207–210**

The doc comment starts a `> 0 = intraday rolling mode:` blockquote but the continuation lines lack `>` markers:

```rust
/// > 0 = intraday rolling mode: warm-up takes this many bars
/// (≈ minutes); entries are evaluated on every bar once warmed up   // ← flagged
/// (no `is_new_day` gate, no `intraday_entries` persistence filter). // ← flagged
/// When set, `intraday_rolling_bars` overrides the window size that  // ← flagged
/// would otherwise come from `lookback` or per-pair `lookback_bars`. // ← flagged
```

Fix: add `>` prefix to each continuation line.

**`collapsible_if` — line 767**

```rust
if trading.intraday_rolling_bars > 0 {
    if let Some(prev_spread) = self.pending_daily_spread.take() {
```

Collapse to `if trading.intraday_rolling_bars > 0 && let Some(...) = ...`.

Both errors appear pre-existing (not in this PR's diff). Confirm: if pre-existing, include a fix here anyway since CI was presumably green on main. If this PR broke them, fix here.

---

### BLOCKER 2: `scripts/topk_sweep.sh:34` passes `--pipeline force`

`force` was removed as a pipeline alias — the runner now calls `process::exit(1)` on unknown profiles. The sweep script is broken:

```bash
"$RUNNER" replay \
    --engine snp500 \
    --start 2026-01-02 --end 2026-04-09 \
    --candidates trading/year2026_candidates_top${N}.json \
    --pipeline force \          ← exits with error code 1
    --bar-cache data/bar_cache_2026 \
```

Change to `--pipeline lab`. Also update two comment-only references:
- `config/metals.toml:177`: `# ... --pipeline force (bypasses pair-picker ADF)` → `--pipeline lab`
- `config/pairs.toml:513`: `# ... relax --pipeline to "force"` → `"lab"`

---

### Non-blocking: stale log literal in `pair_picker_service.rs:94`

```rust
info!(pairs = pairs.len(), path = %path.display(), "wrote active_pairs.json");
```

The `path` field is correct and will show the actual filename, but the string literal still says `active_pairs.json`. In production this logs e.g. `wrote active_pairs.json path=trading/monthly_pairs_202601.json` which is confusing. Change the string to `"wrote pairs file"`.

---

### Non-blocking: stale module doc in `pipeline.rs:12`

```rust
//! writes `active_pairs.json` with passing pairs sorted by score.
```

The filename is no longer meaningful — the function writes to whatever path is passed. Drop the filename from the module doc.

---

### Minor: `autoresearch/dataset.py:4` leftover oracle reference

`Every downstream tool (oracle, pickers, eval) uses this module` — the oracle crate was deleted. Update to remove the oracle reference.

---

## What looks good

- Removal of thompson, lockfile, graph, regime, oracle, and standalone binary is complete and clean
- The `active_pairs.json` legacy fallback removal in commit 95cc59ee is correct
- `aggregate_to_daily` in `alpaca.rs`: NaN/zero guard, RTH filter (13:30–20:00 UTC), BTreeMap-keyed-by-date-string guarantees chronological order, last-tick-wins logic is correct
- Mock server correctly routes `1Min` requests to raw bars → Rust aggregates; Rust always owns the aggregation
- `PairPickerConfig` TOML struct with `Default` impl is the right pattern
- `preserve_input_order` overlay from TOML at the call site in `main.rs` is clean
- Baseline identical at 38 trades / 73.7% / +$3,399 across all commits confirms no regressions

Fix the 2 blockers (clippy + topk_sweep.sh) and CI will pass. Happy to approve once those are addressed.

PR #246, opened 2026-04-13
Add refresh.rs — sync quant-data parquets on runner startup

kumargu:
## Summary

Pure Rust module that syncs quant-data parquet files to latest on every runner startup. No more stale warmup data.

## What it does

On every `openquant-runner` invocation (replay, paper, live), before warmup:

1. Scans each symbol's parquet in `~/quant-data/bars/v2_sp500_2025-2026_1min_adjusted/`
2. Checks contiguity in last 90 trading days — each RTH day (13:30-20:00 UTC) must have ≥50 bars
3. **If corrupt day found**: cleans data from that point onward, re-fetches from Alpaca 1-min IEX API
4. **If contiguous but behind**: fetches only the missing bars incrementally
5. Appends to parquets, deduplicates by timestamp, sorts, writes back

## Implementation

- `engine/crates/runner/src/refresh.rs` — new module (~400 lines)
- Uses `arrow` + `parquet` crates for parquet read/write (new dependencies)
- Uses existing `AlpacaClient.fetch_minute_bars_raw()` for Alpaca fetches
- Schema matches `build_foundation_dataset.py` exactly: `timestamp[us,UTC], open, high, low, close, volume, trade_count, vwap`
- Wired into `main.rs` — called before config load, before warmup

## Tests

7 unit tests with mock parquet data:
- `test_read_write_roundtrip` — write parquet, read back, verify timestamps
- `test_contiguity_good` — 5 full trading days, all pass
- `test_contiguity_corrupt_day` — 1 day with only 20 bars, detected as corrupt
- `test_full_read_write` — full 8-column read/write cycle
- `test_micros_to_date` / `test_micros_to_hm` — timestamp conversion
- `test_alpaca_bars_to_columns` — Alpaca bar format → parquet columns

## Why

Paper trading was starting with stale warmup data (parquets only up to last manual refresh). The rolling z-score window was missing recent days, producing wrong signals. This fix ensures data is always current before trading starts.

## New dependencies

- `arrow = "54"` (with `chrono-tz` feature)
- `parquet = "54"` (with `arrow` feature)


kumargu:
**Code Review — PR #246 (feature/refresh-bars)**

CI is still pending — reviewing the diff now. Overall this is a well-structured addition — the schema matches the Python parquets, the RTH filtering logic is sound, the warn-and-continue error handling is appropriate for a startup check, and the 7 unit tests cover the core paths. However there are **6 blocking issues** that must be fixed before merge.

---

## Blocking issues

### 1. `refresh` runs in replay mode — hits real Alpaca against historical dates

`main.rs` wires `refresh_all` unconditionally, before any run-mode branch. When invoked as `replay --start 2025-10-01 --end 2025-12-31`, it fires `refresh_all` targeting *today's* date and issues live Alpaca API calls. Replay is supposed to be deterministic — bars on disk should not change between runs. This breaks backtest reproducibility: two runs of the same replay can produce different results if bars changed between them.

Fix: gate refresh on `RunMode::Stream(_)` only.

```rust
if let RunMode::Stream(_) = &run_mode {
    let bars_dir = refresh::default_bars_dir();
    // ...
}
```

---

### 2. Deprecated `chrono::DateTime::from_timestamp` — clippy -D warnings will break CI

`refresh.rs` line 194:
```rust
let dt = chrono::DateTime::from_timestamp(secs, 0).unwrap();
```
`chrono::DateTime::from_timestamp` has been `#[deprecated]` since 0.4.27. With `clippy -D warnings` in CI this is a hard failure. Replace with `chrono::DateTime::<chrono::Utc>::from_timestamp(secs, 0).expect("timestamp out of range")`, or sidestep the round-trip entirely:
```rust
let naive = chrono::DateTime::from_timestamp(secs, 0)
```
— actually just use the non-deprecated path directly:
```rust
let dt = chrono::Utc.timestamp_opt(secs, 0).single()
    .expect("timestamp out of range");
```

---

### 3. Non-atomic parquet writes — crash between `File::create` and `writer.close()` corrupts the file permanently

`write_parquet` does `File::create(path)` which truncates the existing file immediately. If the process is killed mid-write (OOM, SIGTERM, power loss), the file is left empty. On the next startup `read_timestamps` returns an empty vec → `Err("empty parquet")` → refresh logs a warning and skips it, never repairs it. The symbol is silently dropped from warmup forever.

Fix: write to a temp file then atomically rename:
```rust
let tmp = path.with_extension("parquet.tmp");
let file = std::fs::File::create(&tmp).map_err(|e| format!("create tmp: {e}"))?;
// ... write ...
writer.close().map_err(|e| format!("close: {e}"))?;
std::fs::rename(&tmp, path).map_err(|e| format!("rename: {e}"))?;
```
On Linux/macOS `rename` is atomic when src and dst are on the same filesystem.

---

### 4. Missing snappy compression — Rust rewrites parquets without compression, breaking parity with Python

`ArrowWriter::try_new(file, schema, None)` uses the default writer properties, which is **uncompressed**. The Python builder uses `compression="snappy"` (`build_foundation_dataset.py:114`). After the first incremental refresh, every touched symbol's parquet grows ~3-4x and no longer matches the files Python produced. Downstream metadata checks and disk-space budgets will be affected silently.

Fix:
```rust
use parquet::file::properties::WriterProperties;
use parquet::basic::Compression;

let props = WriterProperties::builder()
    .set_compression(Compression::SNAPPY)
    .build();
let mut writer = ArrowWriter::try_new(file, schema, Some(props))
    .map_err(|e| format!("writer: {e}"))?;
```

---

### 5. `trade_count` and `vwap` are hardcoded to 0 — silently degrades any downstream consumer

`alpaca_bars_to_columns` always writes `0i64` for `trade_count` and `0.0` for `vwap`. Any code that reads `vwap` from these parquets (pair-picker warmup, engine VWAP signals) gets silent zeros for all freshly-fetched bars. The Alpaca `/v2/stocks/bars` REST response **does include** `vw` (VWAP) and `n` (trade count) — they're just not in `AlpacaBar`.

Fix: add the fields to `AlpacaBar` in `alpaca.rs`:
```rust
pub struct AlpacaBar {
    pub t: String,
    pub o: f64,
    pub h: f64,
    pub l: f64,
    pub c: f64,
    pub v: f64,
    #[serde(default)]
    pub vw: Option<f64>,   // vwap
    #[serde(default)]
    pub n: Option<i64>,    // trade count
}
```
Then in `alpaca_bars_to_columns`:
```rust
vwap.push(bar.vw.unwrap_or(0.0));
trade_count.push(bar.n.unwrap_or(0));
```
Add a test asserting that a bar with `vw: 101.5, n: 200` populates the columns correctly.

---

### 6. `default_bars_dir` hardcodes the dataset version string — breaks when data rolls over

```rust
PathBuf::from(home).join("quant-data/bars/v2_sp500_2025-2026_1min_adjusted")
```

The `v2_sp500_2025-2026_1min_adjusted` suffix will need updating when we roll data to 2026-2027, bump to v3, or run on a different machine. This belongs in config. At minimum expose a `QUANT_DATA_BARS_DIR` env var override, and replace the `/Users/gulshan` fallback (machine-specific, will silently break in CI):

```rust
pub fn default_bars_dir() -> PathBuf {
    if let Ok(dir) = std::env::var("QUANT_DATA_BARS_DIR") {
        return PathBuf::from(dir);
    }
    let home = std::env::var("HOME").unwrap_or_else(|_| String::from("/home/runner"));
    PathBuf::from(home).join("quant-data/bars/v2_sp500_2025-2026_1min_adjusted")
}
```

---

## Non-blocking (address before next PR touching this module)

**A. Triple-read in the corrupt path is wasteful.** When a corrupt day is detected, the parquet is read 3 times: `read_timestamps` → `read_full_parquet` for cleaning → `write_parquet` (cleaned) → `read_full_parquet` again at line 368 to append. The cleaned data is already in memory after the filter step — pass it directly to the append logic instead of re-reading from disk.

**B. `pub mod refresh` should be `mod refresh`.** Binaries have no consumers of their public module API. The `pub` on individual functions inside `refresh.rs` is sufficient for tests. The module declaration in `main.rs` doesn't need to be `pub`.

**C. Sequential fetch over 500 symbols with no concurrency or rate-limit handling.** The loop issues one HTTP call per symbol. For 500 symbols each fetching a 1-7 day gap, this is O(500) serial requests. Alpaca free tier is 200 req/min; a post-holiday gap refresh could take 3+ minutes of blocking startup. `fetch_minute_bars_raw` already accepts `Vec<String>` — batch 10-20 symbols per call to reduce round-trips by an order of magnitude.

**D. No end-to-end test for the corrupt-day repair path.** `test_contiguity_corrupt_day` tests detection only. The highest-risk code path — write parquet with corrupt day → call `refresh_symbol` (with a mock Alpaca client) → verify corrupt rows are removed and the function returns the correct count — has no test. This code mutates production data; it needs coverage.

---

Please fix the 6 blocking items and push. I'll re-review once CI is green.

PR #279, opened 2026-04-23
Fix basket session handling across DST

kumargu:
## Summary
- centralize U.S. equity cash-session handling in a shared New York market-session helper
- remove fixed UTC RTH assumptions from basket live, replay, stream watchdog, refresh, and Alpaca historical aggregation
- add DST-aware tests for session detection and close timestamp construction

## Why
The basket stack was hard-coding `13:30-20:00 UTC` as regular trading hours. That is only correct during daylight saving time. In standard-time months, replay, warmup, live close processing, and stale-symbol monitoring were all using the wrong session window.

## Validation
- `cargo test -p openquant-runner -- --nocapture`

Issue #270, opened 2026-04-23
Review: Basket strategy implementation (M1-M6)


kumargu:
## Overview

Comprehensive review of the basket stat-arb strategy implementation across 9 PRs. This is an end-of-day rebalancing strategy using OU/Bertram thresholds on daily closes.

## PRs to Review

| PR | Milestone | Description |
|----|-----------|-------------|
| #257 | M1 | `basket-picker` crate — OU/AR(1), Bertram thresholds, validator, loader |
| #258 | M2 | `basket-engine` crate — Bertram symmetric state machine |
| #260 | M3 | Portfolio layer + strengthen basket IDs |
| #261 | M4 | Runner integration — `--engine basket` flag, replay harness |
| #262 | M5 | Walk-forward basket parity — matches quant-lab within 2% |
| #264 | Fix | Basket replay parity: dedupe + cost scaling |
| #265 | M6 | Basket live/paper runner with shadow mode |
| #268 | Fix | Default `--engine basket` to universe TOML |
| #269 | Fix | Deep logging pass — stream / basket-engine / basket-live |

## Review Focus Areas

1. **Signal generation**: Are OU/Bertram thresholds computed correctly? Do they match the academic literature?
2. **Entry/exit logic**: Z-score triggers, stop-loss, max-hold — is the state machine correct?
3. **Position sizing**: `aggregate_positions()` → `diff_to_orders()` pipeline — any gaps?
4. **Buying power**: Currently no check — positions can exceed account size
5. **Execution timing**: EOD rebalancing at 19:59 UTC + 2min grace — is this sound?
6. **Warmup**: How many days of history before trading? Is it sufficient?
7. **Numerical correctness**: NaN guards, edge cases, division by zero
8. **Test coverage**: Are there integration tests? Edge case tests?

## Key Files

```
engine/crates/basket-picker/src/
  - lib.rs (OU/AR(1) fitting, Bertram thresholds)
  - bertram.rs (threshold math)
  - validator.rs (basket validation)

engine/crates/basket-engine/src/
  - engine.rs (BasketEngine, on_bars(), state machine)
  - portfolio.rs (aggregate_positions, diff_to_orders)

engine/crates/runner/src/
  - basket_live.rs (live/paper runner, EOD trigger)
  - stream.rs (WebSocket streaming, heartbeats)

config/basket_universe_v1.toml (universe definition)
```

## Context

- Strategy uses daily closes aggregated from 1-min IEX bars
[...]

PR #280, opened 2026-04-23
Fail closed on incomplete basket live data

kumargu:
## Summary
- treat partial Alpaca subscription acknowledgements as fatal so the stream reconnects instead of trading with a subset of symbols
- fail the basket live runner when a weekday session reaches close+grace with no close snapshot or with missing symbols
- exclude today's partial parquet day from live warmup validation
- add stream coverage for the fail-closed subscription path

## Why
The live basket path was still willing to keep running after incomplete subscriptions and partial or empty close snapshots. That is unsafe for a multi-leg EOD strategy because stale or missing legs can silently distort signals and orders. Warmup was also still able to fit against today's incomplete parquet tail on intraday starts.

## Validation
- `cargo test -p openquant-runner -- --nocapture`

### DIFF
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -29,7 +29,7 @@ use basket_engine::{
     PositionIntent, Side,
 };
 use basket_picker::{load_universe, validate, ValidatorConfig};
-use chrono::{DateTime, NaiveDate, Timelike, Utc};
+use chrono::{DateTime, Datelike, NaiveDate, Timelike, Utc, Weekday};
 use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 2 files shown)


kumargu (commented):
## Review — Fail closed on incomplete basket live data

Addresses the fail-open concerns from #270 review (finding 3c). Good direction but needs rebase on #279 first.

### ✅ What's fixed

1. **Partial subscription → Err** — `log_subscription_ack()` now returns `Err` on mismatch, propagates to crash the stream handler. Test coverage added.

2. **Empty close bucket on weekday → fatal** — Weekday with zero buffered closes now returns `Err` instead of silently marking processed. Weekends still mark processed and continue.

3. **Missing symbols at close trigger → fatal** — If any expected symbols are missing from the snapshot, fails with explicit error.

4. **Today's partial parquet excluded from warmup** — `date >= today` filter prevents stale intra-day data from contaminating validation.

### ⚠️ Rebase needed on #279

Both PRs modify `basket_live.rs` and `stream.rs`. More importantly:

- **#280 uses `today = now.date_naive()`** (UTC calendar date)
- **#279 changes this to `market_session::trading_day_utc(now)`** (ET calendar date)

Without the #279 changes, the weekday check (`today.weekday()`) operates on UTC date rather than trading date. Near midnight ET (e.g., 20:00 ET = 01:00 UTC next day), UTC "today" is tomorrow. This could cause:
- Friday session at 20:30 ET to think it's Saturday in UTC
- Skip closing logic because "it's a weekend"

After rebase: use `trading_day_utc(now)` and the weekday check will be correct.

### ✅ CI

Both checks pass (on current base).

### Verdict

**Request changes** — merge #279 first, then rebase this PR to use the timezone-aware session handling. After rebase, this is ready to merge.

kumargu:
@codex review

kumargu:
@codex review

PR #283, opened 2026-04-23
Persist basket live state across restarts

kumargu:
## Summary
- persist basket engine state after each processed session close
- reload state on startup when it matches the frozen fit artifact basket set
- fail closed if the broker already has open positions but there is no matching basket state snapshot
- tighten `BasketEngine::load_state()` validation so malformed snapshots are rejected
- derive a default state path from the fit artifact path

## Why
The live runner was still starting the basket engine flat after every restart even when the broker already had exposure. That breaks the long/short state machine and can produce incorrect flips or liquidation behavior on the next close. This PR makes restart state explicit instead of reconstructing it implicitly.

## Validation

### DIFF
--- a/engine/crates/basket-engine/src/engine.rs
+++ b/engine/crates/basket-engine/src/engine.rs
@@ -347,6 +347,17 @@ impl BasketEngine {
             params.insert(p.basket_id.clone(), p);
         }
 
+        for basket_id in params.keys() {
+            if !snapshot.states.contains_key(basket_id) {
+                return Err(format!("missing runtime state for basket_id '{basket_id}'"));
+            }
+        }
+        for basket_id in snapshot.states.keys() {
+            if !params.contains_key(basket_id) {
+                return Err(format!("runtime state present for unknown basket_id '{basket_id}'"));
+            }
+        }
+
         Ok(Self {
             params,
             states: snapshot.states,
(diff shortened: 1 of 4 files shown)


kumargu (commented):
## Review — Persist basket live state across restarts

Addresses the engine state reconstruction gap I flagged in #270 review (finding 3a: "BasketEngine starts flat on restart even when broker has positions").

### ✅ What's new

1. **State persistence** — engine state saved after each session close to `<fit>.state.json`
2. **State reload** — on startup, loads persisted state if it exists and validates basket set matches fit artifact
3. **Fail-closed gate** — if broker has positions but no state file exists, returns explicit error instead of starting flat
4. **Stricter `load_state()` validation** — rejects snapshots with missing or orphan basket_ids

### ✅ Correctness

The critical safety gate at `basket_live.rs:161-168`:
```rust
if execution.alpaca_mode().is_some() && !state_exists && !current_notionals.is_empty() {
    return Err("open broker positions found but no basket state snapshot exists");
}
```

This prevents the exact bug: engine thinks all baskets are Flat, but Alpaca has exposure → next signal produces incorrect flip/liquidation. Now it fails loud and clear.

State is persisted AFTER `process_session_close()` — if process crashes mid-session, previous day's valid state remains.

### ✅ Design

- Basket ID set validation ensures we don't load stale state from a different universe
- Default path derivation (`<fit>.state.json`) keeps artifacts colocated
- CLI flag `--state-path` allows override for testing/debugging

### ⚠️ CI

No checks have run yet — wait for green before merge.

### Minor

No file locking on state file — concurrent processes could race. Acceptable for single-process deployment; flag if multi-process is planned.

### Verdict

**Ready to merge** once CI passes. This closes the engine state reconstruction gap from #270 and directly addresses finding 3a.

Issue #273, opened 2026-04-23
Basket: buying-power and exposure cap enforcement


kumargu:
## Problem

No buying-power or active-basket cap enforcement. If more than `n_active_baskets` baskets are live, gross exposure exceeds the configured budget.

**Impact**: Can submit orders exceeding account buying power. Observed $340K orders on $100K account in prior run.

## Current behavior
- `aggregate_positions()` allocates `(capital * leverage) / n_active_baskets` per basket
- No clamp if actual active > configured max
- No broker buying-power check or pre-trade margin check

## Expected behavior
- Query Alpaca buying power before order submission
- Scale down proportionally if planned notional > available
- Hard cap on gross exposure
- Reject or warn if too many baskets active

## References
- Codex review finding #3 on #270
- `engine/crates/basket-engine/src/portfolio.rs:30,83`
- Prior session: order rejections due to insufficient buying power

PR #285, opened 2026-04-23
Enforce basket active-position cap

kumargu:
## Summary
- make `aggregate_positions()` reject live books whose active basket count exceeds the configured cap
- fail the basket live session close path instead of quietly building an over-budget target book
- add basket-engine coverage for cap-breach rejection

## Why
The basket portfolio layer sized sleeves as if `n_active_baskets` were enforced, but the engine could hold more active baskets than the configured cap. That silently increases gross exposure beyond the declared budget. This PR turns that into an explicit fail-closed condition.

## Validation

### DIFF
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -84,7 +84,19 @@ pub fn basket_to_legs(params: &BasketParams, position: i8, notional: f64) -> Vec
 pub fn aggregate_positions(
     engine: &BasketEngine,
     config: &PortfolioConfig,
-) -> HashMap<String, f64> {
+) -> Result<HashMap<String, f64>, String> {
+    let active_count = engine
+        .iter_params()
+        .filter_map(|(basket_id, _)| engine.get_state(basket_id))
+        .filter(|s| s.position != 0)
+        .count();
+    if active_count > config.n_active_baskets {
+        return Err(format!(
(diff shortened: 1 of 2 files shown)


kumargu (commented):
## Review — Enforce basket active-position cap

Addresses the `n_active_baskets` enforcement gap from #273 (buying-power and exposure cap).

### ✅ What's fixed

`aggregate_positions()` now fails if active basket count exceeds the configured cap:
```rust
if active_count > config.n_active_baskets {
    return Err("active basket cap exceeded: active={}, configured_cap={}");
}
```

`process_session_close()` propagates the error → entire session fails closed rather than building an over-budget target book.

### ✅ Correctness

- Counts baskets with `position != 0` (actually holding) — correct metric
- Check happens BEFORE target notionals are computed — prevents partial over-allocation
- Fail-closed is the right conservative choice: alerts operator rather than silently partial-trading

### ✅ Tests

`test_aggregate_positions_rejects_cap_breach` — creates 2 active baskets with cap=1, verifies rejection.

### ⚠️ CI

No checks have run yet — wait for green.

### Scope note

This enforces the config-level `n_active_baskets` cap, but doesn't yet query Alpaca buying power (the other half of #273). That can be a follow-up.

### Verdict

**Ready to merge** once CI passes. Partial close of #273.

kumargu (commented):
**Update after rebase:** Active-basket cap enforcement unchanged, rebased on current main. CI pending — ready to merge once green.

PR #287, opened 2026-04-24
[codex] Clean up basket live startup sequence

kumargu:
## What changed
- harden basket startup state loading by restoring saved basket runtime states onto the current frozen fit artifact params instead of trusting persisted params wholesale
- persist `last_processed_trading_day` inside the same basket engine snapshot file so startup catch-up is idempotent with engine state
- run one same-day catch-up close cycle when the runner starts after close+grace on an unprocessed trading day
- require finalized close bars for that catch-up path instead of trusting a partial same-day parquet tail
- add simple startup readiness gates: Alpaca account preflight for paper/live and a first-bar stream health check during RTH

## Why
The live basket runner was improving in pieces, but startup was still too brittle:
- state restore could drift from the current fit artifact contract
- a post-close restart could miss the day entirely or replay it ambiguously
- paper/live startup had no account sanity gate
- the stream path had no minimal proof that live bars were actually flowing before waiting for the next close
- same-day catch-up needed to fail closed on incomplete local close data

This keeps the design close-driven and operationally simple: one command, deterministic startup, fail closed on ambiguity.

## Impact
- basket live can start at any time of day and establish a cleaner state for that trading day
- after-close startup now performs exactly one catch-up processing pass if today is a completed but unprocessed trading session
- that catch-up path only runs from finalized close bars
- paper/live startup fails early on blocked or unusable Alpaca accounts
- intraday startup waits for a first live RTH bar when available, but no longer aborts the whole runner on a transient stream delay

## Validation
- `cargo fmt --all`
- `cargo test -p basket-engine -p openquant-runner -- --nocapture`

## Commit split
1. `Harden basket startup state handling`
2. `Add basket startup readiness gates`
3. `Tighten basket startup persistence and catch-up`
4. `Require finalized close bars for startup catch-up`

PR #288, opened 2026-04-24
[codex] Use share inventory for basket live reconciliation

kumargu:
## What changed
- switch basket live startup reconciliation from mark-to-market notionals to explicit broker share inventory
- convert end-of-day target notionals into target shares at session processing time
- make order generation a pure current-shares to target-shares diff
- keep the target-share conversion fail-closed when close prices are missing or invalid
- add focused tests for target-share rounding and missing-price failure

## Why
The live basket runner was still storing broker state as notionals, which made restart reconciliation approximate and price-dependent. That was the wrong contract for a close-driven equity strategy.

This PR moves the live runner to the cleaner boundary:
- broker state = current shares
- strategy output = target notionals
- session close converts target notionals to target shares using the finalized close snapshot
- orders are the share delta between those two states

## Impact
- startup now reflects actual broker inventory instead of `qty * mark`
- live reconciliation is less sensitive to restart timing and intermediate price moves
- failed orders preserve prior share state rather than overwriting broker state with notional targets
- target-share conversion still fails closed if required close prices are unavailable

## Validation
- `cargo fmt --all`

### DIFF
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -142,12 +142,11 @@ pub struct OrderIntent {
 
 /// Compute orders needed to move from current to target positions.
 ///
-/// Takes current notionals, target notionals, and current prices.
+/// Takes current shares and target shares.
 /// Returns the orders needed to reach target.
 pub fn diff_to_orders(
     current: &HashMap<String, f64>,
     target: &HashMap<String, f64>,
-    prices: &HashMap<String, f64>,
 ) -> Vec<OrderIntent> {
     let mut orders = Vec::new();
 
@@ -157,20 +156,10 @@ pub fn diff_to_orders(
     all_symbols.dedup();
 
     for symbol in all_symbols {
-        let current_notional = current.get(symbol).copied().unwrap_or(0.0);
-        let target_notional = target.get(symbol).copied().unwrap_or(0.0);
-        let delta = target_notional - current_notional;
-
-        if delta.abs() < 1.0 {
-            continue; // Skip tiny deltas
-        }
-
-        let price = match prices.get(symbol) {
-            Some(&p) if p.is_finite() && p > 0.0 => p,
-            _ => continue, // Skip if no valid price
-        };
-
-        let qty = (delta.abs() / price).round() as u32;
+        let current_shares = current.get(symbol).copied().unwrap_or(0.0);
+        let target_shares = target.get(symbol).copied().unwrap_or(0.0);
+        let delta = target_shares - current_shares;
+        let qty = delta.abs().round() as u32;
         if qty == 0 {
             continue;
         }
@@ -268,20 +257,16 @@ mod tests {
         target.insert("AMD".to_string(), 3000.0);
         target.insert("NVDA".to_string(), 2000.0);
 
-        let mut prices: HashMap<String, f64> = HashMap::new();
-        prices.insert("AMD".to_string(), 100.0);
-        prices.insert("NVDA".to_string(), 200.0);
-
-        let orders = diff_to_orders(&current, &target, &prices);
+        let orders = diff_to_orders(&current, &target);
 
         assert_eq!(orders.len(), 2);
-        // AMD: 3000 - 5000 = -2000, sell 20 shares
+        // AMD: 3000 - 5000 = -2000 shares, sell 2000 shares
         let amd_order = orders.iter().find(|o| o.symbol == "AMD").unwrap();
         assert_eq!(amd_order.side, Side::Sell);
-        assert_eq!(amd_order.qty, 20);
-        // NVDA: 2000 - 0 = 2000, buy 10 shares
+        assert_eq!(amd_order.qty, 2000);
+        // NVDA: 2000 - 0 = 2000 shares, buy 2000 shares
         let nvda_order = orders.iter().find(|o| o.symbol == "NVDA").unwrap();
         assert_eq!(nvda_order.side, Side::Buy);
-        assert_eq!(nvda_order.qty, 10);
+        assert_eq!(nvda_order.qty, 2000);
     }
 }
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -200,35 +200,27 @@ pub async fn run_basket_live(
         fresh
     };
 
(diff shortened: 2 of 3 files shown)


kumargu:
## Review: share-based live reconciliation

**Overall:** This is the right contract. Tracking broker state as shares and keeping the strategy in notionals — with the conversion happening once at session close — eliminates the restart-vs-price-move race that the old mark-to-market seeding had. The math is cleaner, the diff is deterministic, and the failure modes are now explicit (fail-closed on missing close, preserve prior shares on failed orders).

### Correctness

**1. Alpaca qty sign — verified OK.**
`get_positions` parses `qty` as `f64`. Alpaca's `/v2/positions` endpoint returns shorts with a negative `qty` string (e.g. `\"-100\"`), so `p.qty.parse::<f64>()` preserves sign. The new `seed_current_shares_from_alpaca` drops `avg_entry` entirely, which is correct — we no longer need a price to reconstruct a notional.

**2. Fail-closed on missing close — correct.**
`target_shares_from_notionals` collects missing symbols, sorts, and returns `Err`. This bubbles up through `process_session_close` and halts the session. Good. The new test covers both the rounding and the failure path.

**3. Reconciliation loop simplification — correct.**
The old \"no order emitted + price valid\" branch is gone because:
- target_shares already filters out no-price symbols (fails closed), so
- any symbol still in `target_shares` is guaranteed to have been convertible
- the only remaining drift source is execution failure (handled by `orders_by_symbol` branch)

The union iteration (`current ∪ target`) is preserved, so symbols dropped from the target (fully-exited baskets) are still cleared. Good.

**4. Dollar threshold → share threshold — cleaner.**
Old `diff_to_orders` had `if delta.abs() < 1.0 { continue }` (skip sub-$1 deltas). New version has `delta.abs().round() as u32 == 0` (skip sub-half-share deltas). Since `target_shares_from_notionals` only inserts if `shares.abs() >= 1.0`, small targets never enter the diff. No regression.

### Design / tech debt

**5. Dead code introduced.** `previous_trading_day` and `latest_completed_trading_day_utc` were just added in #287 for startup reconciliation. With the reference-close load now removed, they're only called from tests within `market_session.rs` — hence the new `#[cfg_attr(not(test), allow(dead_code))]` attributes.

Per `CLAUDE.md`:
> Remove dead code aggressively: If a function is defined but never called, delete it or wire it in. Dead code misleads readers and rots.

Please either delete both functions (and their tests) or move them under `#[cfg(test)]` if the intent is to keep them as documented behavior. A `#[cfg_attr(not(test), allow(dead_code))]` on a prod function is the wrong signal — the function ships in the binary but is unreachable.

**6. Minor: `avg_entry` unused.** `seed_current_shares_from_alpaca` now destructures `(qty, _avg_entry)`. If no other caller needs `avg_entry_price`, consider simplifying `get_positions` to return just `HashMap<String, f64>` (qty). Not blocking for this PR.

### CI

Both checks currently pending — will verify once they complete.

### Verdict

LGTM on the core refactor once the dead-code cleanup is addressed. The contract change (broker=shares, strategy=notionals, close=conversion, orders=share-delta) is the right boundary for a close-driven equity strategy.

kumargu:
## Review of commit 24108380 (non-basket filter)

**LGTM.** Tight, narrowly-scoped fix.

**What it does:** Filters `seed_current_shares_from_alpaca` output to the basket universe, so broker positions from other strategies sharing the same Alpaca account don't leak into `current_shares`.

**Why it's necessary:** Without this filter, a non-basket position (e.g. AAPL held by a different strategy) would enter `current_shares` with qty=X but never appear in `target_shares`. The reconciliation loop would then emit a close-order for AAPL on first session — aggressively liquidating another strategy's position. Hard failure mode.

**Implementation is correct:**
- `allowed` is built from the basket universe symbols (`symbols` passed from caller)
- `filter_map` drops unknown symbols and records them for logging
- `warn!` emits `ignored_symbols.len()` plus a bounded 10-symbol sample (good — not unbounded log spam)
- Sort before logging for determinism

**Minor nit (non-blocking):** The sample truncation uses `.take(10).collect::<Vec<_>>()` — this allocates a `Vec<&String>` just for logging. For a typical basket universe (tens of symbols) this is negligible, and the sort+take pattern is clear. Not worth changing.

**CI:** ✅ All checks passing.

**Dead-code concern from earlier review still stands** for `previous_trading_day` / `latest_completed_trading_day_utc`. Otherwise approved.

kumargu:
## Review of commit 276cadd3 (refresh before diff)

**This is a meaningful design change.** The previous approach tracked per-order success locally and updated `current_shares` only for executed orders. The new approach:
1. Refreshes `current_shares` from Alpaca **before** computing `diff_to_orders`
2. Submits orders but **doesn't** mutate `current_shares` afterward (paper/live)
3. Relies on the next session's startup refresh to reconcile actual fills

### Pros
- **Simpler data flow:** No local tracking of which orders succeeded. The broker is the source of truth — always.
- **Drift auto-corrects:** If an order partially fills or gets rejected, the next session's refresh picks up the real state. No accumulating phantom drift.
- **Cleaner noop mode:** Shadow mode directly assigns `*current_shares = target_shares`, which is deterministic across sessions.

### Cons / risks
**1. Intra-session gap:** Between submitting orders and the next session's refresh, `current_shares` reflects **stale** broker state (the pre-diff snapshot). If the code ever computes a second diff in the same session (which currently it doesn't), it would re-emit the same orders.

**Mitigation:** The bar loop only calls `process_session_close` once per trading day (`processed_sessions.insert(today)` guard). So this gap doesn't matter in practice. But the mental model changed — `current_shares` is no longer "current" after order submission until next session startup.

**2. Refresh latency:** `seed_current_shares_from_alpaca` is async and hits the broker. This moves Alpaca latency into the hot path of `process_session_close` (previously it was only at startup). For a close-driven strategy this is fine — it runs once per day. Flagging for awareness.

**3. Contract documentation drift:** The earlier reconciliation loop had detailed comments explaining the union iteration and per-symbol update semantics. Now that's deleted. The replacement comment ("next session refresh will reconcile actual fills") is accurate but terse. Consider adding a one-liner explaining why we don't update `current_shares` after order submission.

### Verdict
**LGTM — cleaner design.** The broker-as-source-of-truth model is easier to reason about than the local per-order tracking. The intra-session staleness is non-issue given the once-per-day close processing. CI Rust tests pending; will confirm once green.

PR #289, opened 2026-04-24
[codex] Enforce basket portfolio admission caps

kumargu:
## What changed
- add a portfolio admission step in `basket-engine` that ranks non-flat baskets by `|last_z|` and admits only the top `n_active_baskets`
- generate target notionals only from the admitted basket set instead of sizing every active basket regardless of cap
- expose admission results so the live runner can log admitted vs excluded baskets
- add an explicit gross-notional hard cap check against `capital * leverage` before order generation
- add coverage for active-basket cap enforcement in `basket-engine`

## Why
The live basket path still had a core portfolio bug: if more baskets were active than `n_active_baskets`, it sized all of them anyway. That could exceed the configured portfolio budget even though the config implied a hard cap.

This PR makes the portfolio layer enforce the cap directly instead of relying on implicit operator discipline.

## Impact
- only the strongest active baskets are admitted into the target portfolio when the engine has more non-flat baskets than allowed
- excluded baskets are logged explicitly in the live runner
- gross target exposure now fails closed if it exceeds the configured capital/leverage budget

## Validation
- `cargo fmt --all`

### DIFF
--- a/engine/crates/basket-engine/src/engine.rs
+++ b/engine/crates/basket-engine/src/engine.rs
@@ -404,6 +404,15 @@ impl BasketEngine {
         self.params.iter()
     }
 
+    /// Flatten a set of baskets so portfolio admission and engine state stay aligned.
+    pub fn flatten_baskets(&mut self, basket_ids: &[String]) {
+        for basket_id in basket_ids {
+            if let Some(state) = self.states.get_mut(basket_id) {
+                state.flatten();
+            }
+        }
+    }
+
     fn validate_snapshot(snapshot: &EngineSnapshot) -> Result<(), String> {
         let mut params = HashMap::new();
         for p in &snapshot.params {
--- a/engine/crates/basket-engine/src/lib.rs
+++ b/engine/crates/basket-engine/src/lib.rs
@@ -11,8 +11,8 @@ mod state;
 pub use engine::{BasketEngine, BasketParams, EngineSnapshot};
 pub use intent::{PositionIntent, TransitionReason};
 pub use portfolio::{
-    aggregate_positions, basket_to_legs, diff_to_orders, LegNotional, OrderIntent, OrderReason,
-    PortfolioConfig, Side,
+    aggregate_positions, basket_to_legs, diff_to_orders, plan_portfolio, LegNotional, OrderIntent,
+    OrderReason, PortfolioConfig, PortfolioPlan, Side,
 };
 pub use state::BasketState;
 
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -1,6 +1,6 @@
 //! Portfolio aggregation and order generation.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use serde::{Deserialize, Serialize};
 
@@ -43,6 +43,19 @@ pub struct LegNotional {
     pub notional: f64,
 }
 
+/// Portfolio admission result after applying active-basket caps.
+#[derive(Debug, Clone)]
+pub struct PortfolioPlan {
+    /// Symbol-level target notionals for admitted baskets only.
+    pub symbol_notionals: HashMap<String, f64>,
+    /// Basket ids admitted into the portfolio target set.
+    pub selected_baskets: Vec<String>,
+    /// Active basket ids excluded by the cap.
+    pub excluded_baskets: Vec<String>,
+    /// Number of non-flat baskets seen before applying the cap.
+    pub active_baskets: usize,
+}
+
 /// Compute leg notionals for a basket position.
 ///
 /// For a basket with target and N peers:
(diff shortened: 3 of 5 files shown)


kumargu:
## Review: Portfolio admission caps

**Overall:** This is the right fix. The old code relied on the operator to not enter more baskets than configured — a dangerous implicit contract when the engine can autonomously flip into any basket that hits z-threshold. This PR makes the cap explicit and enforced.

### Correctness

**1. Admission ranking by |last_z| — good signal choice.**
Higher |z| = stronger dislocation = higher conviction. Sorting by descending |z| admits the most extreme signals first. The tiebreaker on `basket_id` is deterministic. No hidden randomness.

**2. Cap enforcement before notional aggregation — correct ordering.**
The old `aggregate_positions` iterated all non-flat baskets and sized them all. Now `plan_portfolio`:
1. Collects all active (non-flat) baskets with |z|
2. Sorts by |z| descending
3. Takes top `n_active_baskets`
4. Only then aggregates notionals for selected baskets

This guarantees gross exposure can't exceed `capital * leverage * (n/n)` = `capital * leverage` (the configured budget).

**3. Gross-notional hard cap — defense in depth.**
The check `gross_notional > gross_cap + 1.0` catches pathological edge cases (rounding, overlapping legs, etc.). Fails closed with `Err`. The `+ 1.0` epsilon avoids floating-point false positives.

**4. Test coverage — good.**
`test_plan_portfolio_enforces_active_basket_cap`:
- Creates 2-basket engine
- Sets cap to 1
- Asserts `selected=1`, `excluded=1`
- Asserts `gross == notional_per_basket` (i.e., exactly 1 basket's worth)

**5. `aggregate_positions` preserved for backward compat.**
Now delegates to `plan_portfolio(...).symbol_notionals`. Existing callers (if any outside runner) continue to work.

### Design

**6. `PortfolioPlan` struct — clean API.**
Caller gets:
- `symbol_notionals`: what to trade
- `selected_baskets`: who's in
- `excluded_baskets`: who's out + why (for logging)
- `active_baskets`: total before cap

The runner logs all of this. Observability greatly improved.

**7. Double-iteration over `iter_params()` — minor inefficiency.**
First pass collects active baskets + |z|, second pass aggregates legs for selected set. Could be merged, but clarity > micro-optimization at this frequency (once per day). Fine.

### CI

✅ Both checks passing.

### Minor nits (non-blocking)

- **Logging:** `excluded_sample = ?plan.excluded_baskets.iter().take(10)` — same truncation pattern as #288. Good consistency.
- **Sort stability:** `partial_cmp` with `NaN` handling via `unwrap_or(Equal)` is correct. `last_z` is `Option<f64>`, unwrapped to 0.0 if None, so no NaN expected. Belt-and-suspenders is fine.

### Verdict

**LGTM.** Ready to merge. This closes the portfolio-budget loophole that could've caused runaway exposure.

kumargu:
## Review of commits d2f306cf + e62d3de7

### Commit d2f306cf: Flatten excluded baskets

**Intent:** When a basket is excluded from the portfolio due to the active-basket cap, flatten it so the engine state reflects "no position."

**Implementation:**
- `BasketState::flatten()` — clears `position`, `entry_date`, `entry_spread` while preserving diagnostics (`last_z`, `spread_history`, etc.)
- `BasketEngine::flatten_baskets(&mut self, basket_ids)` — bulk-flattens a list of baskets
- `process_session_close` calls `engine.flatten_baskets(&plan.excluded_baskets)` after admission

**Correctness concern — this is a forced exit without orders.**

When a basket is excluded from the portfolio (e.g., a 3rd basket when cap=2), this code flattens its engine state *without* emitting close orders. That means:
1. Broker may still hold the position (if the basket was entered in a prior session)
2. Engine state now says "flat" but broker state says "holding"
3. Next session's `seed_current_shares_from_alpaca` pulls real broker shares → but engine doesn't generate close orders because target=0 matches engine's view

**Risk:** Orphaned broker positions. The excluded basket's legs stay open on Alpaca indefinitely.

**Suggested fix:** Either:
a) Emit explicit close orders for excluded baskets before flattening, OR
b) Let `diff_to_orders` handle it naturally — keep excluded baskets out of `target_notionals` (which you do) but DON'T flatten the engine state. Next diff would compute `target=0` vs `current=X` → close orders.

Actually, wait — let me re-read the flow...

The broker refresh happens **before** `plan_portfolio`:
```rust
*current_shares = seed_current_shares_from_alpaca(...);
let plan = plan_portfolio(...);
engine.flatten_baskets(&plan.excluded_baskets);
let target_shares = ...;
let orders = diff_to_orders(current_shares, &target_shares);
```

So `current_shares` has the real broker state. `target_shares` excludes the now-flattened baskets (they contribute 0 to target_notionals). Therefore `diff_to_orders` sees:
- `current` = broker shares (includes excluded basket's legs)
- `target` = 0 for those symbols (unless another basket also holds them)
→ Close orders emitted ✅

**I retract the concern.** The ordering is correct. The flatten happens after target computation is locked, but **before** the diff uses `target_shares`. And `target_shares` is computed from `target_notionals`, which is computed from the **pre-flatten** engine state via `plan_portfolio`. Actually no — `plan_portfolio` already excludes them...

Let me trace again more carefully:
1. `plan = plan_portfolio(engine, ...)` — computes `symbol_notionals` using only **selected** baskets (excluded baskets contribute nothing)
2. `engine.flatten_baskets(&plan.excluded_baskets)` — mutates engine state for excluded baskets
3. `target_shares = target_shares_from_notionals(&target_notionals, closes)` — converts plan's notionals to shares (excludes excluded baskets)
4. `orders = diff_to_orders(current_shares, &target_shares)` — broker vs target → generates close orders for legs held by excluded baskets

**Confirmed correct.** Step 1 already excludes them from notionals. Step 2's flatten is for engine-state consistency (so next session's `plan_portfolio` doesn't re-see them as active). Close orders emerge naturally from step 4.

---

### Commit e62d3de7: Validate nonzero active-basket cap

**Implementation:**
- `PortfolioConfig::validate()` checks capital > 0, leverage > 0, n_active_baskets >= 1
- Called in `run_basket_live` at startup → fails fast on bad config

**Correct and straightforward.** Division by zero on `notional_per_basket()` would've produced `inf` before; now it fails closed with a clear message.

**Test coverage:** `test_portfolio_config_rejects_zero_active_baskets` confirms the validation.

---

### CI

Pending — will confirm when green.

### Verdict

**LGTM.** The flatten logic is correctly sequenced. Excluded baskets:
1. Don't contribute to target notionals
2. Get engine-state flattened (so they don't persist as "active" across sessions)
3. Generate close orders via the natural broker-vs-target diff

Config validation is defense-in-depth against operator error.

PR #290, opened 2026-04-24
[codex] Check basket order set against buying power

kumargu:
## What changed
- add a fresh order-set affordability check in the basket live runner after admission and share-delta generation
- fetch Alpaca account state again at session close for paper/live execution
- fail closed if the concrete basket order set gross notional exceeds current buying power
- add focused test coverage for invalid / nonpositive buying power parsing

## Why
The live runner already had startup account preflight and portfolio-level caps, but it still did not validate the actual session-close basket orders against current broker buying power.

This PR adds that missing broker-aware gate on the concrete order set before any orders are submitted.

## Impact
- paper/live basket execution now fails closed if the generated order set is larger than current broker buying power
- the check runs on the actual admitted order set, not just on startup account state
- logs now show the order-set gross notional versus current buying power when the check passes

## Validation
- `cargo fmt --all`

### DIFF
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -113,24 +113,38 @@ async fn check_order_set_affordability(
     alpaca: &AlpacaClient,
     mode: ExecutionMode,
     date: NaiveDate,
+    current_shares: &HashMap<String, f64>,
+    target_shares: &HashMap<String, f64>,
     orders: &[OrderIntent],
     closes: &HashMap<String, f64>,
 ) -> Result<(), String> {
     let account = alpaca.get_account(mode).await?;
     let buying_power = parse_buying_power(&account)?;
-    let order_gross: f64 = orders
+    let current_gross: f64 = current_shares
+        .iter()
+        .filter_map(|(symbol, shares)| closes.get(symbol).map(|p| shares.abs() * p))
+        .sum();
+    let target_gross: f64 = target_shares
+        .iter()
+        .filter_map(|(symbol, shares)| closes.get(symbol).map(|p| shares.abs() * p))
+        .sum();
+    let incremental_gross = (target_gross - current_gross).max(0.0);
+    let order_turnover: f64 = orders
         .iter()
         .filter_map(|o| closes.get(&o.symbol).map(|p| p * o.qty as f64))
         .sum();
-    if order_gross > buying_power + 1.0 {
+    if incremental_gross > buying_power + 1.0 {
         return Err(format!(
-            "order set gross notional {:.2} exceeds Alpaca buying power {:.2} on {}",
-            order_gross, buying_power, date
+            "incremental gross exposure {:.2} exceeds Alpaca buying power {:.2} on {}",
+            incremental_gross, buying_power, date
         ));
     }
     info!(
         date = %date,
-        order_gross_notional = %format!("{:.0}", order_gross),
+        current_gross_notional = %format!("{:.0}", current_gross),
+        target_gross_notional = %format!("{:.0}", target_gross),
+        incremental_gross_notional = %format!("{:.0}", incremental_gross),
+        order_turnover_notional = %format!("{:.0}", order_turnover),
         buying_power = %format!("{:.0}", buying_power),
         "order-set affordability check passed"
     );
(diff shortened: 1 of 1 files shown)


kumargu:
## Review: Order-set buying power check

**Overall:** Good defense-in-depth. The portfolio admission cap guards against over-allocation at the strategy layer; this check guards against actually exceeding broker liquidity at execution time.

### Correctness

**1. Check placement — correct.**
Called inside `process_session_close` after orders are generated but before `place_order` loop:
```rust
Some(mode) => {
    check_order_set_affordability(alpaca, mode, date, &orders, closes).await?;
    // ... place orders
}
```
If it fails, no orders are submitted. Clean fail-closed.

**2. Gross computation logic.**
```rust
let order_gross: f64 = orders
    .iter()
    .filter_map(|o| closes.get(&o.symbol).map(|p| p * o.qty as f64))
    .sum();
```
Sums `price * qty` for each order. Uses close prices from the session snapshot. Missing prices are silently skipped via `filter_map`.

**Minor concern:** If a close price is missing for an order symbol, that order's notional won't be counted toward `order_gross`, making the check more permissive. In practice this shouldn't happen — `target_shares_from_notionals` already fails closed on missing prices, so by the time we reach here every order symbol should have a close. But if future code paths bypass that gate, this could under-count. Consider logging a warning if `closes.get(&o.symbol)` returns None, or asserting.

**3. `+ 1.0` epsilon — consistent with portfolio cap.**
Matches the `gross_notional > gross_cap + 1.0` pattern from PR #289. Avoids float false-positives.

**4. `parse_buying_power` validation.**
- Parses string → f64 (Alpaca returns buying_power as string)
- Rejects non-finite or non-positive values
- Test covers `buying_power: "0"` → `Err("not positive")`

Good. Edge cases (negative margin, account issues) won't slip through as valid.

**5. Fresh account fetch at session close.**
```rust
let account = alpaca.get_account(mode).await?;
```
Fetches current account state rather than relying on startup snapshot. Buying power can change intraday (fills, deposits, margin calls). Correct to re-check.

### Design

**6. Async call in hot path.**
Another Alpaca API call in `process_session_close` (joins the broker-share refresh from #288). For a once-per-day close-driven strategy, this latency is negligible. Flagging for awareness — if this ever becomes per-bar, it'd need batching.

**7. No noop-mode check.**
The `check_order_set_affordability` call is inside `Some(mode) =>`, so noop mode (which has no Alpaca mode) skips it. Correct — shadow mode doesn't need buying-power validation.

### CI

✅ Both checks passing.

### Verdict

**LGTM.** Ready to merge. This closes the gap between "portfolio says it's within budget" and "broker says you can afford it."

kumargu:
## Review of commits a295950a + 2220789b (incremental + side-aware exposure)

### What changed

**1. Incremental exposure instead of gross order notional.**
The original check compared total order notional vs buying power. This fails legitimate rotations: selling $10K of AMD to buy $10K of NVDA is $20K of orders but $0 incremental exposure — the proceeds from the close fund the open.

Now the check computes:
```rust
let incremental_long = (target_long_gross - current_long_gross).max(0.0);
let incremental_short = (target_short_gross - current_short_gross).max(0.0);
let incremental_exposure = incremental_long + incremental_short;
```
This measures *net new* capital required on each side. A self-financing rotation has `incremental_exposure = 0`.

**2. Sell-before-buy order sequencing.**
```rust
ordered_refs.sort_by_key(|o| match o.side {
    Side::Sell => 0_u8,
    Side::Buy => 1_u8,
});
```
Executes sells before buys so proceeds are available. This is critical for the incremental-exposure model to be realistic — if you buy first, you may not have the liquidity that the check assumed.

### Correctness

**3. `gross_by_side` helper — correct.**
Iterates share-map, multiplies by close price, accumulates into `long_gross` (positive notional) or `short_gross` (abs of negative notional). Missing prices are skipped.

**4. Test coverage — good.**
- `test_incremental_gross_logic_allows_self_financing_rotation_shape`: current=100 AMD @ $100 → target=50 NVDA @ $200 → same $10K long gross → incremental=0. ✅
- `test_incremental_exposure_counts_long_to_short_reversal`: current=100 AMD long → target=100 AMD short → new short exposure = $10K. ✅

**5. Logging verbosity — appropriate.**
Now logs both sides' current/target gross, incremental components, order turnover, and buying power. Rich diagnostics for post-mortem.

### Design consideration

**6. Shorts require margin, not buying power.**
`incremental_short` measures new short exposure, but Alpaca's `buying_power` is cash + margin available for *longs*. Short positions require `short_margin` or `reg_t_buying_power` depending on account type.

For a simple equity margin account, Alpaca's `buying_power` is generally 2× equity for longs but shorts have separate locate/margin requirements. The check as written may be too permissive (allowing large shorts when `buying_power` is high) or too strict (blocking shorts that don't need buying power at all).

**Practical impact:** For a portfolio running 50/50 long-short with modest leverage, this is unlikely to bite. But it's worth noting that `buying_power` isn't the precise gate for short exposure. A future refinement could use `regt_buying_power`, `daytrading_buying_power`, or `initial_margin` depending on account type.

**Not blocking** — the check is defense-in-depth, and Alpaca will reject orders that truly exceed account limits. This is a "better safe than sorry" gate, not the source of truth.

### CI

Pending — will confirm when green.

### Verdict

**LGTM.** The incremental model is correct for a self-financing rotation strategy. Sell-before-buy ordering ensures liquidity is available when buys execute. The margin/buying_power nuance is a known limitation but not a blocker.

PR #295, opened 2026-04-24
Clean up runner CLI params: remove dead Engine methods (#293)

### LINKED ISSUE #293: Simplify and prune runner CLI params
## Problem

\`StreamArgs\` in \`engine/crates/runner/src/main.rs\` has accumulated a mix of params that are engine-specific, redundant, or documented as unused. The CLI has gotten hard to reason about: every \`paper\`/\`live\` invocation has to think about which flags apply to which engine.

## Concrete smells

- **Dead default**: \`Engine::Basket.config_path()\` returns \`\"config/basket.toml\"\` with inline comment \`// Not used; basket uses universe TOML\`. It's a lie that exists in the type system.
- **Universal flags that are engine-specific**:
  - \`--universe\`, \`--bars-dir\`, \`--fit-artifact\`, \`--state-path\`, \`--execution\` are all basket-only but sit on \`StreamArgs\`.
  - \`--candidates\` is snp500/metals-only.
- **Override flags that may all be dead**: \`--config\`, \`--pipeline\` are \"override the engine default\" escape hatches. \`Engine::pipeline()\` even says \"All engines use 'lab'\" — so the override likely hasn't been exercised in ages.
- **Redundancy**: \`--data-dir\` and \`--trading-dir\` both default to fixed paths; unclear when anyone passes them.
- **Naming drift**: \`execution = \"noop\"\` is the default, but the word \"noop\" is confusing. \`shadow\` or \`log-only\` is clearer. Same goes for \`paper\` subcommand + \`--execution paper\` — the double-nesting is weird.

## Proposal

1. **Audit every flag** — grep \`git log\` + CI scripts for each \`--foo\` usage. Kill anything that's never set.
2. **Factor per-engine args into sub-structs** — something like \`BasketArgs\` with \`universe/bars_dir/fit_artifact/state_path/execution\`, selected by \`--engine basket\`. Don't put basket-only fields on a struct shared with snp500/metals.
3. **Delete \`Engine::Basket.config_path()\`** — it returns a path that's never read. Remove the method or return \`Option\`.
4. **Clarify \`--execution\`** — rename \`noop\` → \`shadow\` (or \`log-only\`). Pick one idiom and document it.
5. **Minimum-viable paper invocation** — the happy path should be \`openquant-runner paper --engine basket\` with \`--execution paper\`. No other flags. Anything more required is a smell.
6. **Integration test**: \`openquant-runner --help\` output snapshotted and kept under N lines, so the CLI doesn't regrow this mass.

## Acceptance

- Happy-path paper command for basket is ≤ 2 flags (\`--engine basket --execution paper\`)
- No unused CLI flag remains
- \`--help\` fits on a laptop screen for each subcommand
- Per-engine fields live on per-engine arg structs, not a shared blob

kumargu:
Closes #293 (first pass — smallest, safest cleanup).

## Summary

- Delete three Engine helper methods — \`config_path\`, \`candidates_path\`, \`pipeline\` — that were only ever called from \`resolve_engine\`, which is short-circuited for \`--engine basket\` in \`main()\`. Their \`Basket\` arms returned paths that were never read (one literally had \`// Not used; basket uses universe TOML\` admitting as much).
- Inline the pair-engine defaults in \`resolve_engine\` with an explicit \`unreachable!()\` for \`Basket\`, documenting why that variant never reaches this function.
- Delete \`DEFAULT_CONFIG\` constant — \`resolve_engine\` always populates config, so the \`unwrap_or_else\` fallback in \`run()\` was dead.
- Return \`(PathBuf, ...)\` from \`resolve_engine\` instead of \`(Option<PathBuf>, ...)\` and tighten \`run()\`'s signature accordingly.
- Update \`Engine\` and \`StreamArgs\` doc comments to honestly describe which flags belong to which engine.

No CLI behavior changes. \`--help\` output for pair-engine flags now says \"pair engines only; basket ignores this\" instead of the misleading \"Override config file (default: selected by --engine)\" for basket.

This is the smallest, safest slice of #293. Factoring \`StreamArgs\` into per-engine sub-structs and renaming \`--execution noop → shadow\` are follow-ups.

## Test plan

- [x] \`cargo test -p openquant-runner\` — pass (3 pre-existing tests ignored)
- [x] \`./openquant-runner paper --help\` — verified flags say the right thing now

### DIFF
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -149,6 +149,17 @@ pub fn plan_portfolio(engine: &BasketEngine, config: &PortfolioConfig) -> Portfo
         .skip(config.n_active_baskets)
         .map(|(basket_id, _)| basket_id.clone())
         .collect();
+    debug_assert!(
+        selected_baskets.len() <= config.n_active_baskets,
+        "admission cap violated: selected {} > cap {}",
+        selected_baskets.len(),
+        config.n_active_baskets
+    );
+    debug_assert_eq!(
+        selected_baskets.len() + excluded_baskets.len(),
+        active_baskets,
+        "selected + excluded must equal active baskets"
+    );
     let selected: HashSet<&str> = selected_baskets.iter().map(|s| s.as_str()).collect();
 
     for (basket_id, params) in engine.iter_params() {
@@ -233,11 +244,20 @@ pub fn diff_to_orders(
     for symbol in all_symbols {
         let current_shares = current.get(symbol).copied().unwrap_or(0.0);
         let target_shares = target.get(symbol).copied().unwrap_or(0.0);
+        debug_assert!(
+            current_shares.is_finite() && target_shares.is_finite(),
+            "non-finite shares for {symbol}: current={current_shares} target={target_shares}"
+        );
         let delta = target_shares - current_shares;
+        debug_assert!(
+            delta.is_finite(),
+            "non-finite share delta for {symbol}: {delta}"
+        );
         let qty = delta.abs().round() as u32;
         if qty == 0 {
             continue;
         }
+        debug_assert!(qty > 0, "zero qty slipped past the filter for {symbol}");
 
         let side = if delta > 0.0 { Side::Buy } else { Side::Sell };
 
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -218,6 +218,15 @@ fn summarize_orders_by_side(
     (buy_count, sell_count, buy_notional, sell_notional)
 }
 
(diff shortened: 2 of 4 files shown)


kumargu:
I don’t think this is merge-safe yet as a `#293` “smallest, safest CLI cleanup” against current `main`.

The diff still includes unrelated basket/live-path changes in `engine/crates/runner/src/basket_live.rs`, `engine/crates/basket-engine/src/portfolio.rs`, and `engine/crates/runner/src/market_session.rs`. The most important one is the new post-submission reconciliation block in `process_session_close()` that sleeps for 30 seconds and refetches broker positions after order submission. That is runtime behavior, not CLI cleanup.

Because of that, merging `#295` would silently ship basket execution-path changes under a PR whose title/body say there is no behavior change. I would rebase/split this PR so the actual `#293` slice is just the `main.rs` CLI cleanup. Once those basket/live changes are removed from the branch, the CLI cleanup itself looks reasonable.

Issue #294, opened 2026-04-24
Replay must exercise the live code path (not a separate backtest)


kumargu:
## Problem

\`replay --engine basket\` runs a **separate** walk-forward backtest (\`basket_runner::run_basket_replay\`) with its own P&L formula (\`pos_{t-1} × (spread_t − spread_{t-1})\`). It does **not** call \`basket_live::process_session_close\`.

This defeats the entire reason replay was written in Rust.

**The purpose of Rust replay was to exercise the exact live code path with noop execution against Alpaca**, so that:
- Every bug fix in the live path gets coverage in replay
- Backtest P&L reflects what the live engine would actually have done
- Regressions in execution logic (portfolio caps, buying power, reconciliation, share diffing, order emission) get caught before they lose real money

Today we have two engines masquerading under one name:

| | Replay (basket_runner) | Live (basket_live) |
|---|---|---|
| Portfolio cap (\`gross > cap\` Err) | ❌ bypassed | ✅ enforced |
| Buying power check | ❌ bypassed | ✅ enforced |
| Share-inventory reconciliation | ❌ bypassed | ✅ enforced |
| Post-submit reconciliation alert | ❌ bypassed | ✅ enforced |
| \`PortfolioConfig\` applied | ❌ ignored | ✅ applied |
| P&L formula | Spread-delta approximation | Real share × price fills |

The \"does the signal work?\" question the backtest was built for is also the least interesting one at this stage — **execution quality dominates P&L at our current scale**. Yesterday's $1,237 loss was entirely execution-driven (60/108 orders rejected → lopsided book). A replay that doesn't exercise the code that failed is useless for validating the fix.

## Proposal

1. **Delete \`basket_runner::run_basket_replay\`** (or demote it to a \`statarb-backtest\` dev tool under \`crates/research/\`). Stop calling it from the \`replay\` subcommand.
2. **Rewire \`replay --engine basket\`** to:
   - Load parquet bars for the requested \`--start\`/\`--end\` range
   - Construct a \`NoopAlpacaClient\` (or inject a mock) that accepts orders but doesn't hit the network
   - Call \`run_basket_live\` with that client + a synthetic clock that advances one session at a time
   - Emit the same structured logs we get in paper/live + a per-session P&L synthesized from simulated fills
3. **One code path, parameterized by clock and broker.** \`process_session_close\` never needs to know it's replay vs live.
[...]

PR #296, opened 2026-04-24
Extract Broker trait for order placement (#294a)

kumargu:
First of three PRs decomposing #294 (replay must exercise the live code path).

This PR is **pure refactor: zero runtime behavior change**. Introduces a \`Broker\` trait and implements it for the existing \`AlpacaClient\`.

## Why

\`basket_live.rs\`'s helpers (\`preflight_account_check\`, \`check_order_set_affordability\`, \`seed_current_shares_from_alpaca\`, \`process_session_close\`) all took \`&AlpacaClient\` directly, hard-wiring them to \`paper-api.alpaca.markets\`. For replay to exercise the live code path against a simulated broker (as #294 requires), these helpers must be polymorphic over the broker backend.

## Changes

- \`crates/runner/src/broker.rs\` new — \`trait Broker\` with 3 async methods: \`place_order\`, \`get_positions\`, \`get_account\`. Covers the broker-action surface only; historical bar fetching stays on \`AlpacaClient\` (that's data plane, not execution plane).
- \`impl Broker for AlpacaClient\` just delegates to the existing inherent methods.
- The four helpers in \`basket_live.rs\` swap \`alpaca: &AlpacaClient\` → \`broker: &impl Broker\` and rename parameters accordingly.

\`run_basket_live\` itself still takes \`&AlpacaClient\` — it needs the websocket API key/secret for \`stream::start_bar_stream\`. Abstracting the bar source is PR #294b; that's when \`AlpacaClient\` can disappear from the top-level signature too. Follow-up PR adds the \`SimulatedBroker\` impl.

## Test plan

- [x] \`cargo test -p openquant-runner\` — pass
- [x] Paper run against paper-api continues unaffected (this refactor is type-level only; the currently-running paper process uses the pre-refactor binary)

Part of #294.

### DIFF
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -149,6 +149,17 @@ pub fn plan_portfolio(engine: &BasketEngine, config: &PortfolioConfig) -> Portfo
         .skip(config.n_active_baskets)
         .map(|(basket_id, _)| basket_id.clone())
         .collect();
+    debug_assert!(
+        selected_baskets.len() <= config.n_active_baskets,
+        "admission cap violated: selected {} > cap {}",
+        selected_baskets.len(),
+        config.n_active_baskets
+    );
+    debug_assert_eq!(
+        selected_baskets.len() + excluded_baskets.len(),
+        active_baskets,
+        "selected + excluded must equal active baskets"
+    );
     let selected: HashSet<&str> = selected_baskets.iter().map(|s| s.as_str()).collect();
 
     for (basket_id, params) in engine.iter_params() {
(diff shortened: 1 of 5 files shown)


kumargu:
The Broker-trait direction is fine, but I would not merge this branch as-is because it is not isolated to the trait refactor.

The diff to current `main` still carries unrelated basket/live changes from lower branches: startup/order observability changes, market-session helper removal, portfolio debug assertions, and the post-submission broker reconciliation block inside `process_session_close()`. So this is not actually a pure type-level refactor against the current codebase.

I’d rebase this branch onto current `main` (or land/split the lower runtime changes first) so the review surface is just `broker.rs` plus the signature swaps in `basket_live.rs` / `main.rs`. Right now the scope is too mixed to trust the claimed zero-behavior-change contract.

PR #297, opened 2026-04-24
Extract BarSource trait for the bar feed (#294b)

kumargu:
Second of three PRs decomposing #294 (replay must exercise the live code path).

**Stacks on #296** (#294a — Broker trait). Reviewers should land #296 first; this PR targets \`refactor/broker-trait-294a\` and will be retargeted to \`main\` once #296 merges.

Pure refactor, zero runtime behavior change.

## Why

\`run_basket_live\` reached into \`&AlpacaClient\` to grab \`api_key\` and \`api_secret\` for \`stream::start_bar_stream\`. That coupled the live loop to the concrete Alpaca WebSocket — replay can't feed bars through this path without a seam.

## Changes

- New \`crates/runner/src/bar_source.rs\` with \`trait BarSource\` — single async method: \`start(&self, symbols) -> mpsc::Receiver<StreamBar>\`. The \`StreamBar\` type stays in the existing \`stream\` module so the consumer loop in \`run_basket_live\` is unchanged.
- \`AlpacaBarSource\` struct holds the WebSocket creds and implements \`BarSource\` by delegating to \`stream::start_bar_stream\`.
- \`run_basket_live\` gains a \`bar_source: &impl BarSource\` parameter. It no longer needs \`AlpacaClient\`'s \`api_key\`/\`api_secret\` directly.
- Caller (\`main::run_basket_stream\`) constructs \`AlpacaBarSource\` from the \`AlpacaClient\`'s credentials and passes both down.

Together with the \`Broker\` trait from #294a, the basket live loop now depends only on abstractions — the concrete \`AlpacaClient\` is only materialized at the edge in \`main.rs\`. **PR #294c** will add \`SimulatedBroker\` + \`ParquetBarSource\` and wire them through \`replay --engine basket\`.

## Test plan

- [x] \`cargo test -p openquant-runner\` — pass
- [x] Running paper process unaffected (uses pre-refactor binary)

Part of #294.

### DIFF
--- a/engine/crates/basket-engine/src/portfolio.rs
+++ b/engine/crates/basket-engine/src/portfolio.rs
@@ -149,6 +149,17 @@ pub fn plan_portfolio(engine: &BasketEngine, config: &PortfolioConfig) -> Portfo
         .skip(config.n_active_baskets)
         .map(|(basket_id, _)| basket_id.clone())
         .collect();
+    debug_assert!(
+        selected_baskets.len() <= config.n_active_baskets,
+        "admission cap violated: selected {} > cap {}",
+        selected_baskets.len(),
+        config.n_active_baskets
+    );
+    debug_assert_eq!(
+        selected_baskets.len() + excluded_baskets.len(),
+        active_baskets,
+        "selected + excluded must equal active baskets"
+    );
     let selected: HashSet<&str> = selected_baskets.iter().map(|s| s.as_str()).collect();
 
     for (basket_id, params) in engine.iter_params() {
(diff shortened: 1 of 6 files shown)


kumargu (commented):
Incrementally, the BarSource extraction itself looks sound: pushing the bar feed behind a small trait is the right seam if replay is going to exercise the live loop.

I do not have a material finding on the `#297` delta relative to `#296`. The real caution is stack hygiene: because `#297` sits on `#296`, and `#296` still needs a rebase/scope cleanup against current `main`, I would treat this PR as conditionally OK only after the lower branch is cleaned up and the stack is retargeted.

Issue #300, opened 2026-04-25
[#294c-3] Hardening: failure injection, delete duplicate replay math, preserve parity TSV


kumargu:
Parent: #294 · Depends on: #299

## Why this exists

#299 makes replay *run* through the live code path. This issue closes the loop by:

1. **Hardening** the new path — a deterministic in-process `SimulatedBroker` reconciles with itself, so after #299 we can claim "the reconciliation branch executed" but not "the reconciliation logic is validated." Failure-injection modes fix that.
2. **Deleting the duplicate execution math** in `basket_runner::run_basket_replay` — the whole point of #294.
3. **Preserving the parity TSV harness** as a thin wrapper over the new replay output, so we don't lose quant-lab baseline comparison while we transition.

### Modify `simulated_broker.rs` — failure injection

Add config-driven failure modes:

```rust
pub struct SimulatedBrokerConfig {
    pub slippage_bps: f64,
    pub reject_rate: f64,             // probability of rejecting a place_order
    pub partial_fill_rate: f64,       // probability of filling 60-90% instead of 100%
    pub stale_position_rate: f64,     // probability of returning a stale position snapshot
    pub seed: u64,                    // deterministic for tests
}
```

- Reject path: `place_order` returns `Err("buying power exceeded" | "insufficient liquidity")` matching the shapes Alpaca emits, so the `error!("ORDER FAILED")` branch in `basket_live` runs.
- Partial fill: `place_order` returns `AlpacaOrder { filled_qty: ceil(qty * 0.6..0.9), status: "partially_filled" }`.
- Stale position: `get_positions` returns the snapshot from one bar ago instead of the current state. This is what makes the post-submit reconciliation alert ("BROKER DIVERGENCE") actually fire and lets us assert on it.
- Default config: all rates 0.0 → identical to #299 behavior. Tests opt in.

### Add — golden-snapshot integration test

`engine/crates/runner/tests/replay_basket_golden.rs`:

- Replay a fixed multi-day window (e.g., `2024-07-01..2024-07-15`) with `slippage_bps=0`, all failure rates 0.
- Capture the emitted `OrderIntent` stream + per-session P&L + final positions/cash.
- Compare against a checked-in golden file (`tests/fixtures/replay_basket_golden.json`).
- A second test runs the same window with `reject_rate=0.1, seed=42` and asserts the BROKER DIVERGENCE error fires at least once and the engine recovers cleanly the next session.

[...]

Issue #299, opened 2026-04-25
[#294c-2] Wire replay --engine basket through run_basket_live (SimulatedBroker + ParquetBarSource)


kumargu:
Parent: #294 · Depends on: #298

## Why this exists

`#298` lands the missing seams (`Clock`, `SessionTrigger`). With those in place, `replay --engine basket` can call `run_basket_live` directly with replay-shaped impls and no `if replay {}` branches anywhere.

This is the PR that actually makes replay exercise the live code path.

### Add

- `engine/crates/runner/src/parquet_bar_source.rs` — `impl BarSource`
  - Constructor: `(bars_dir: PathBuf, start: NaiveDate, end: NaiveDate, symbols: Vec<String>)`.
  - `start()` opens per-symbol parquets, walks them via a min-heap merge in chronological order, and pushes `StreamBar` records into the same `mpsc::Sender<StreamBar>` channel the WebSocket uses.
  - Channel closes when all parquets drained → caller exits naturally.
  - Maintains `latest_close: Arc<RwLock<HashMap<String, f64>>>` shared with `SimulatedBroker` (updated on each bar emit).

- `engine/crates/runner/src/simulated_broker.rs` — `impl Broker`
  - Holds:
    - `positions: Arc<RwLock<HashMap<String, (f64, f64)>>>` (qty, avg_entry_price)
    - `cash: Arc<RwLock<f64>>` (init from `PortfolioConfig.capital`)
    - `latest_close: Arc<RwLock<HashMap<String, f64>>>` (shared with `ParquetBarSource`)
    - `slippage_bps: f64` (config knob, default 0)
  - **Replay fill contract**: `place_order` fills at `latest_close[sym] × (1 ± slippage_bps/1e4)` — i.e., **the completed close bar that generated the signal** — and returns a synthetic `AlpacaOrder { status: "filled", filled_qty: qty, filled_avg_price: ... }`.
  - `get_positions` clones the map.
  - `get_account` returns `equity = cash + Σ(qty × latest_close[sym])`, `buying_power = equity × leverage`.
  - **No failure injection in this PR** — happy-path fills only. Rejection / partial-fill / stale-position modes land in #294c-3.

- `engine/crates/runner/src/replay_clock.rs` — `impl Clock` + `impl SessionTrigger`
  - `BarDrivenClock` reads "now" off the latest emitted bar timestamp.
[...]

Issue #298, opened 2026-04-25
[#294c-1] Inject Clock and SessionTrigger into run_basket_live (no behavior change)


kumargu:
Parent: #294 · Replaces part of original #294c plan (post-Codex review).

## Why this exists

The original `#294` plan assumed replay was a 2-seam swap (`Broker` + `BarSource`). Codex pointed out that's not enough: today `run_basket_live` drives session-close cadence off a real `tokio::time::interval(30s)`, and "now" off `Utc::now()`. If `ParquetBarSource` drains a year of bars in 5 seconds, no close ever fires.

So the live path needs two more injection points before replay can hook in:

1. **`Clock`** — what time is it (live: `SystemClock`; replay: derived from the latest bar timestamp).
2. **`SessionTrigger`** — when do we run the close cycle (live: 30s wall-clock interval inspecting `clock.now()` against the session close + grace; replay: bar-driven, fires when a bar's timestamp crosses the next session boundary).

## Scope

Pure refactor. **No behavior change for paper/live.** No replay code lands in this PR.

### Add

- `engine/crates/runner/src/clock.rs`
  ```rust
  pub trait Clock: Send + Sync {
      fn now(&self) -> chrono::DateTime<chrono::Utc>;
  }
  pub struct SystemClock;
  impl Clock for SystemClock { /* Utc::now() */ }
  ```

- `engine/crates/runner/src/session_trigger.rs`
  ```rust
  pub trait SessionTrigger: Send + Sync {
      /// Yields each time a session-close event should fire. Returns the
      /// trading date that just closed.
      async fn next(&mut self) -> Option<chrono::NaiveDate>;
  }
  pub struct IntervalSessionTrigger { /* 30s interval, current behavior */ }
  ```

### Modify

- `engine/crates/runner/src/basket_live.rs`
  - `run_basket_live` accepts `&impl Clock` and `&mut impl SessionTrigger`.
  - Replace the existing `tokio::time::interval(30s)` close-detection loop with `session_trigger.next().await`.
  - Replace `Utc::now()` calls in heartbeat / catch-up / past-close detection with `clock.now()`.
  - Keep all per-bar logic, all engine calls, all logging fields exactly the same.

- `engine/crates/runner/src/main.rs`
  - `live` and `paper` construct `SystemClock` + `IntervalSessionTrigger` and pass them in. No CLI changes.

### No changes

- `Broker` / `BarSource` traits — untouched.
- `basket_runner.rs` — untouched.

[...]

PR #301, opened 2026-04-25
[#294c-1] Inject Clock and SessionTrigger into run_basket_live

kumargu:
## Summary

First of three PRs implementing #294 (replay through the live code path). Codex's review (https://github.com/kumargu/OpenQuant/issues/294#issuecomment-4318488895) flagged that replay needs more than the two seams we already have (\`Broker\`, \`BarSource\`) — today \`run_basket_live\` drives session-close cadence off a real \`tokio::time::interval(30s)\`, and \"now\" off \`Utc::now()\`. If \`ParquetBarSource\` drains a year of bars in seconds, no close ever fires.

This PR adds the two missing injection points so replay (#299) can swap them out.

### Changes

- **\`clock.rs\`** — \`Clock\` trait + \`SystemClock\` impl. Replaces the three direct \`Utc::now()\` calls inside \`run_basket_live\` and its heartbeat block. The two helper functions (\`load_warmup_closes\`, \`load_daily_closes_with_timestamps\`) used by the one-shot \`freeze-basket-fits\` tool are intentionally left as-is — they're not in the live runtime call graph.
- **\`session_trigger.rs\`** — \`SessionTrigger\` trait + \`IntervalSessionTrigger\` impl. The trigger owns 30s polling + dedup of yielded dates. \`basket_live\`'s \`processed_sessions\` set is kept as the **persisted-state** dedup (defends against restart after a session was already processed and its state was saved); the trigger's \`last_yielded\` is the **scheduling** dedup. Two distinct concerns, two distinct sets of state.
- **\`basket_live.rs\`** — \`run_basket_live\` now takes \`&impl Clock\` and \`&mut impl SessionTrigger\`. The select-loop \`_ = tick.tick()\` arm is now \`today = session_trigger.next() => { ... }\`. Heartbeat (60s observability) ticker stays untouched.
- **\`main.rs\`** — live/paper construct \`SystemClock\` + \`IntervalSessionTrigger::new(SystemClock, 2)\`. CLI surface unchanged.

### What's NOT in this PR

- \`BarDrivenClock\` / \`BarDrivenSessionTrigger\` (those are #299)
- \`SimulatedBroker\` / \`ParquetBarSource\` (#299)
- Deleting \`basket_runner::run_basket_replay\` (#300)

### Behavior

No paper/live behavior change. The 30s polling cadence, 2-min grace period, and dedup semantics are preserved exactly.

## Test plan

- [x] New unit test \`session_trigger::tests::dedupes_same_date\` verifies \`IntervalSessionTrigger\` yields a date once and refuses to re-yield while still on the same date, then fires when the clock advances to the next trading day
- [ ] Smoke test: run \`paper --engine basket\` for a few minutes and confirm log stream is structurally identical to pre-#298 (TBD on next paper window — out of scope for this PR)

### DIFF
--- a/engine/crates/runner/Cargo.toml
+++ b/engine/crates/runner/Cargo.toml
@@ -31,3 +31,4 @@ parquet = { version = "54", features = ["arrow"] }
 
 [dev-dependencies]
 tempfile = "3"
+tokio = { version = "1", features = ["full", "test-util"] }
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -37,7 +37,9 @@ use tracing::{debug, error, info, warn};
 use crate::alpaca::ExecutionMode;
 use crate::bar_source::BarSource;
 use crate::broker::Broker;
+use crate::clock::Clock;
 use crate::market_session;
+use crate::session_trigger::SessionTrigger;
 use crate::stream;
 
 /// Execution mode for basket live/paper.
@@ -271,6 +273,8 @@ fn classify_startup_phase(
 pub async fn run_basket_live(
     broker: &impl Broker,
     bar_source: &impl BarSource,
+    clock: &impl Clock,
+    session_trigger: &mut impl SessionTrigger,
     universe_path: &Path,
     fit_artifact_path: &Path,
     state_path: &Path,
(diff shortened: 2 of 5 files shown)


kumargu (commented):
No material findings on `#301`.

I reviewed the `Clock` / `SessionTrigger` refactor against current `main` and the live-path behavior still looks preserved: startup uses injected time, the close cycle is now delegated to a scheduler seam, and the persisted-state dedup remains separate from scheduler dedup.

Residual risk to keep in mind for the next stacked PRs:
- the close-grace value is now duplicated between `basket_live.rs` and `main.rs`, so `#299` should centralize or thread that value rather than relying on comments to keep them in sync
- replay will still need a bar-driven trigger, not just a bar-driven clock, which this PR correctly sets up but does not yet solve

I also ran the new trigger unit test locally:
- `cargo test -p openquant-runner session_trigger::tests::dedupes_same_date -- --nocapture`

It passed.

PR #302, opened 2026-04-25
[#294c-2] Wire replay --engine basket through run_basket_live (delete legacy walk-forward)

### LINKED ISSUE #299: [#294c-2] Wire replay --engine basket through run_basket_live (SimulatedBroker + ParquetBarSource)

kumargu:
## Summary

Closes #299. Refs #294, #298.

`replay --engine basket` now goes through the exact same \`run_basket_live()\` function that paper/live use. The diff is which trait impls get passed in — no \`if replay {}\` branches anywhere in \`basket_live.rs\`.

Per the user's call, this PR also **deletes the legacy walk-forward path entirely** rather than keeping it behind a flag. \`basket_runner.rs\` and the \`pos_{t-1} × (spread_t − spread_{t-1})\` closed-form P&L formula are gone. Parity TSV restoration as a thin wrapper over the new replay output is deferred to #300.

## What's new

- **\`simulated_broker.rs\` (\`SimulatedBroker\`, \`Broker\` impl)** — in-process. Fills at latest emitted close ± configurable \`slippage_bps\`. Tracks positions, cash, equity. Runs a buying-power gate shaped like Alpaca's so the \`error!(\"ORDER FAILED\")\` branch in \`basket_live\` gets coverage even on the happy path. Returns \`AlpacaOrder { status: \"filled\" }\`. **No failure-injection modes** — those land in #300 deliberately so the reconciliation tests there can exercise the divergence branch on demand.

- **\`parquet_bar_source.rs\` (\`ParquetBarSource\`, \`BarSource\` impl)** — reads per-symbol parquets in \`bars_dir\`, merges them in chronological order via min-heap, emits \`StreamBar\`s on an mpsc channel (buffer=1) so basket_live consumes them in lockstep. After each session's bars are drained, signals the \`BarDrivenSessionTrigger\`. Maintains a shared closes map that \`SimulatedBroker\` reads to compute fill prices.

- **\`replay_clock.rs\`** — \`BarDrivenClock\` (returns latest emitted bar's OPEN time via watch channel) and \`BarDrivenSessionTrigger\` (mpsc; yields \`None\` when the bar source is exhausted = basket_live's replay-exit signal).

## Trait API change

\`SessionTrigger::next\` now returns \`Option<NaiveDate>\`. \`None\` = source exhausted. Live's \`IntervalSessionTrigger\` never returns \`None\`. Updated existing test + \`basket_live.rs\` consumer accordingly.

## Replay fill contract (frozen here)

> Fills happen at the *latest known close* for the symbol — the close from the bar that triggered the session-close cycle. Plus configurable one-sided slippage (default 0 bps). No partial fills, no rejections (failure injection is #300).

This matches \"MOC executes at the official close\" used by paper, so replay vs paper P&L stays comparable.

## Replay state isolation

\`--state-path\` defaults to \`<data-dir>/replay/<universe-stem>.state.json\`. Never reads or writes the live default \`.state.json\`. New CLI flags: \`--capital\`, \`--leverage\`, \`--n-active-baskets\`, \`--slippage-bps\`, \`--state-path\`, \`--fit-artifact\`. Removed: \`--baseline\`, \`--tsv-out\`.

## What's deleted

- \`engine/crates/runner/src/basket_runner.rs\` — entire file, including \`run_basket_replay\`, \`process_session_close_replay\`, the spread-delta P&L formula, the parity TSV writer, and the baseline JSON loader.
- \`mod basket_runner;\` in \`main.rs\`.
- The whole \`if a.engine.is_basket() { ... legacy walk-forward block ... }\` branch in the \`replay\` command handler.

## Test plan

- [x] \`cargo test --workspace\` green (39 runner unit tests + 6 doctests)
- [x] 5 new \`SimulatedBroker\` unit tests cover: zero-slippage fill, one-sided slippage (10 bps buy), buying-power rejection, missing-close error, position close resets cost basis
- [ ] **End-to-end smoke test on real parquets** — pending. The \`run_basket_live\` call path needs \`config/basket_universe_v1.fits.json\` to exist locally. I'll run a short \`replay --engine basket --start 2024-07-01 --end 2024-07-08\` after this PR is up to confirm behavior end to end. If anything breaks, follow-up commit on this branch.
- [ ] Golden-snapshot integration test — deferred to #300 per the issue plan.

## What's NOT in this PR (per the #299 spec)

- Failure injection modes for \`SimulatedBroker\` (rejection, partial fill, stale position) → #300
- Parity TSV harness as thin wrapper over new replay output → #300
- Golden-snapshot integration test → #300


kumargu:
Review findings on `#302`:

1. `SimulatedBroker`'s buying-power gate is materially wrong for multi-symbol books, so replay can both under-reject and over-reject orders compared with the intended live-path semantics. In `place_order()` the projected gross is computed as `sum(abs(qty_existing)) * price_of_current_order + notional_of_current_order` ([simulated_broker.rs:132-139](https://github.com/kumargu/OpenQuant/pull/302/files#diff-0d0d7de3d6d398ca9e2f6d3920857f605ad42109b5df8f013e34e8a93f26865cR132-R139)). That applies the *current symbol's* fill price to every existing position, regardless of what symbols are actually held. Example: if the book already holds 10 shares of a $1,000 name and you place a $10 order in a cheap name, `projected_gross` uses `$10` for the existing expensive position and badly understates exposure. The reverse can also falsely reject legitimate orders. Since one of the explicit goals here is to exercise the buying-power path through replay, this needs to compute projected gross from the actual per-symbol close map (or directly project the post-trade position book) rather than a single-price rough bound.

2. The new default replay state path is isolated correctly, but the code never creates its parent directory, so first-run replay will fail on the first state save if `<data-dir>/replay/` does not already exist. `run_basket_replay_live_path()` defaults `state_path` to `data_dir/replay/<universe-stem>.state.json` ([main.rs:1250-1258](https://github.com/kumargu/OpenQuant/pull/302/files#diff-b6d2b3fe06693a355b9d21b291eab0c434c9833833bba0b79c85fd31db1c695fR1250-R1258)), but `BasketEngine::save_state_with_day()` just calls `fs::write(path, ...)` with no `create_dir_all(parent)` ([engine.rs on main still does this](https://github.com/kumargu/OpenQuant/blob/main/engine/crates/basket-engine/src/engine.rs#L338-L350)). So a clean checkout that has `data/` but not `data/replay/` will get all the way into the replay loop and then fail at the first persisted session close. The replay entrypoint should create the parent directory up front or the engine save path should do it generically.

kumargu:
Both findings addressed in d37dfad3:

**1. Buying-power gate** — replaced the single-price approximation with a per-symbol projection:
- Each existing position valued at its own close from the shared closes map
- Symbol being traded valued at its post-trade qty × fill price
- New positions (not yet in the book) added explicitly

Two regression tests:
- `buying_power_uses_each_symbols_own_close` — small order in cheap name passes when an expensive position is near cap; large order correctly rejects
- `closing_a_position_does_not_increase_projected_gross` — closes reduce gross instead of adding notional

**2. Replay state dir** — added `fs::create_dir_all` in `run_basket_replay_live_path` for the state path's parent before calling `run_basket_live`. First-session-close save now succeeds in a clean checkout.

All 7 SimulatedBroker tests pass; clippy + fmt clean.

kumargu:
One remaining material finding after the latest fix commit:

`#302` now isolates replay state to a different default path, but it still does **not** start replay from fresh state if that file already exists. `run_basket_replay_live_path()` builds an isolated `state_path` under `<data-dir>/replay/...` and then calls `run_basket_live()` unchanged ([main.rs:1250-1258](https://github.com/kumargu/OpenQuant/pull/302/files#diff-b6d2b3fe06693a355b9d21b291eab0c434c9833833bba0b79c85fd31db1c695fR1250-R1258), [main.rs:1358-1370](https://github.com/kumargu/OpenQuant/pull/302/files#diff-b6d2b3fe06693a355b9d21b291eab0c434c9833833bba0b79c85fd31db1c695fR1358-R1370)), but `run_basket_live()` still does its normal `if state_exists { load_snapshot(...) ... }` restore path ([basket_live.rs on main around the startup block](https://github.com/kumargu/OpenQuant/blob/main/engine/crates/runner/src/basket_live.rs#L341-L366)).

So the first replay run is isolated, but the *second* replay run over the same universe/data-dir will silently resume from the prior replay snapshot instead of starting flat. That breaks replay determinism and contradicts the intended “fresh isolated state” contract for backtests. A user re-running `replay --engine basket --start X --end Y` should not get path-dependent results from an old `.state.json` unless resume semantics are explicitly requested.

I would fix this one of two ways before merge:
- default replay to deleting/ignoring any existing state file and start from empty engine + empty simulated broker state
- or add an explicit `--resume-state` style flag, with fresh-start as the default for replay

Right now the code has isolation, but not freshness.

kumargu:
Addressed in 69d39790:

By default, replay now **deletes any existing state file** at the resolved `state_path` before calling `run_basket_live`. So the second replay of the same universe starts from empty engine state + empty SimulatedBroker (which is always constructed fresh anyway). Determinism across re-runs preserved.

New `--resume-state` flag (default false) lets you opt back in to loading a prior snapshot, for debugging mid-replay state.

```
# default: fresh start
replay --engine basket --start 2024-07-01 --end 2024-07-15

# opt in to resume:
replay --engine basket --start 2024-07-01 --end 2024-07-15 --resume-state
```

clippy + fmt clean.

kumargu (commented):
Rechecked after the latest two fix commits.

The three material findings I had on `#302` are now addressed:
- buying-power projection now values existing positions using their own closes instead of the current order's price
- replay creates the isolated state directory before first save
- replay now starts fresh by default, with `--resume-state` as explicit opt-in

I reran the new simulated-broker regression tests locally:
- `cargo test -p openquant-runner buying_power_uses_each_symbols_own_close -- --nocapture`
- `cargo test -p openquant-runner closing_a_position_does_not_increase_projected_gross -- --nocapture`
- `cargo test -p openquant-runner errors_on_missing_close -- --nocapture`

They passed.

At this point I do not have a remaining material correctness finding on `#302`. The main residual risk is operational rather than architectural: this still needs the promised end-to-end replay smoke run on real parquets before merge, because the PR deletes the old basket replay path entirely.

kumargu:
## Smoke run (1-week window) ✅

```
replay --engine basket --start 2024-07-01 --end 2024-07-08 --capital 10000
```

**Result:**
- 52,773 bars processed across 5 trading sessions
- 43 orders placed across the window, **all filled**, 0 failed
- `accepted_orders=4 failed_orders=0` per session-close cycle
- Engine state persisted to `data/replay/basket_universe_v1.state.json` (isolated path)
- Clean exit via `REPLAY EXHAUSTED — SHUTDOWN`
- No ERRORs, no panics

**Final snapshot:**
```
initial_capital = 10000.00
final_cash       = 10514.87
final_equity     = 9901.05
final_pnl        = -98.95   (-0.99%)
positions        = 27
```

The full live code path (`run_basket_live → process_session_close → diff_to_orders → broker.place_order → post-submit reconciliation`) ran through the SimulatedBroker / ParquetBarSource / BarDrivenClock / BarDrivenSessionTrigger swap with no `if replay {}` branches in basket_live.

Ready to merge.

Issue #256, opened 2026-04-21
[M5] replay parity — verify Rust basket engine P&L matches quant-lab


kumargu:
Parent: #251
Depends on: #255 (runner integration)

## Scope (REVISED — was paper trading, now replay parity)

**Paper trading deferred to later tonight / follow-up issue.**

M5 now validates that the Rust basket engine produces P&L matching quant-lab's backtest within tolerance. This is the critical correctness gate before any live/paper execution.

## What to implement

**Run the replay:**
```bash
openquant-runner replay --engine basket \
  --universe engine/config/universe/basket_universe_v1.toml \
  --start 2024-07-01 --end 2026-04-13 \
  --pnl-out pnl_rust.csv
```

**Compare vs quant-lab baseline:**
- Reference: point Sharpe 2.80, CI [1.26, 3.18], +145.18% cum return, -13.65% max DD
- Rust output must match within tolerance:
  - Cum return: ±2%
  - Sharpe: within [2.70, 2.90]
  - Max drawdown: ±1%
  - Trade count: ±5%

**Where mismatches point to bugs:**
If P&L diverges, file targeted issues for each discrepancy:
- Spread computation mismatch
- OU fitting divergence
- Bertram threshold calculation
- Entry/exit timing
- Cost accounting
- Position sizing

## Expected workflow

1. Coder runs replay, outputs CSV
2. Reviewer compares CSV vs quant-lab output
3. If parity achieved (within tolerance): merge M5, proceed to paper trading (new milestone)
4. If mismatch: file M6+, keep M5 loop open, reviewer comments on each issue with:
   - Observed metric
   - Expected metric
   - Suggested investigation area

## Acceptance

- [ ] `replay --engine basket` runs end-to-end without panics
- [ ] CSV output with daily gross/net P&L, cumulative
- [ ] Cum return within ±2% of quant-lab baseline (+145.18%)
- [ ] Sharpe within [2.70, 2.90]
- [ ] All bug issues from parity analysis filed and closed

## Out of scope (deferred to M6+)

- Paper trading via websocket (original M5 scope)
- Live Alpaca execution
- Real-money trading

cc #251

kumargu:
## Scope Change — M5 is now replay parity (not paper trading)

Per user direction: M5 scope has been revised from "paper trading live websocket" to "replay parity verification against quant-lab baseline".

[...]

PR #304, opened 2026-04-25
[#294c-3] Replay hardening: failure injection + parity TSV + EOD equity

### LINKED ISSUE #300: [#294c-3] Hardening: failure injection, delete duplicate replay math, preserve parity TSV

kumargu:
## Summary

Closes #300. Refs #294, #298, #299. Final PR in the replay-through-live-path epic.

Adds three pieces on top of the #299 replay path:
1. Failure-injection modes in `SimulatedBroker`
2. End-of-day equity tracking via a new `Broker::record_eod` hook
3. A new replay report module + TSV writer

### `SimulatedBroker` failure injection

Three deterministically-seeded modes — all default to OFF so #299 happy-path replay behavior is unchanged:

- **\`reject_rate\`** — per-order probability of returning a \"buying power exceeded\"-shaped \`Err\`. Exercises basket_live's \`error!(\"ORDER FAILED\")\` branch.
- **\`partial_fill_rate\`** — per-order probability of filling 60-90% of the requested qty (status=\"partially_filled\"). Drives BROKER DIVERGENCE on the post-submit reconciliation path.
- **\`stale_position_rate\`** — per-\`get_positions\` probability of returning the previous snapshot. Also drives BROKER DIVERGENCE.

Single \`StdRng\` seeded from \`SimulatedBrokerConfig.seed\` → deterministic outcomes.

### End-of-day equity tracking

\`Broker::record_eod(date)\` (default noop, real op on \`SimulatedBroker\`). Called from \`basket_live\` after every \`process_session_close\`. Builds a \`BTreeMap<NaiveDate, f64>\` of mark-to-market equity. Pure mark-to-market on simulated fills.

### \`replay_report.rs\` module

New replay report — **not** drop-in compatible with the deleted \`basket_runner.rs\` walk-forward TSV. Conventions are documented in the module header:

- \`ann_return\` is linearized (\`daily_mean × 252\`)
- \`max_drawdown\` is reported as a positive fraction \`(peak − equity) / peak\`
- TSV emits one section: \`date\\tequity\\tdaily_pnl_dollar\`

API: \`ReplayStats\`, \`compute_stats\`, \`write_tsv\`. No \`BaselineJson\` / \`print_parity_comparison\` / \`--baseline\` flag — replay no longer gates against quant-lab baselines.

### CLI surface (\`replay --engine basket\`)

\`\`\`
--reject-rate <f64>           [default 0]
--partial-fill-rate <f64>     [default 0]
--stale-position-rate <f64>   [default 0]
--failure-seed <u64>          [default 0]
--report-tsv <PATH>           [default off]
\`\`\`

### Failure semantics

A failed replay run now exits non-zero **before** report generation, so a non-zero exit never leaves behind a misleading TSV. The post-mortem broker snapshot still gets logged for debugging.

## Smoke verification (1-week window, all failure modes on)

\`\`\`
replay --engine basket --start 2024-07-01 --end 2024-07-08 \\
  --capital 10000 \\
  --reject-rate 0.1 --partial-fill-rate 0.1 --stale-position-rate 0.05 \\
  --failure-seed 42 --report-tsv data/replay/report_smoke.tsv
\`\`\`

Verified:
- Multiple \`ORDER FAILED ... error=\"buying power exceeded: simulated rejection for ...\"\` → reject path works
- \`BROKER DIVERGENCE: actual gross differs from target by >10% target=38400 actual=33885 divergence_pct=11.8 accepted_orders=23 failed_orders=4\` → reconciliation path works
- Engine recovered cleanly the next session
- \`REPLAY EXHAUSTED — SHUTDOWN\` clean exit
- \`REPLAY PORTFOLIO STATS cum_return=-0.0078 sharpe=-2.103 max_dd=+0.0223\`
- TSV written with correct format

## Test plan

- [x] **52 runner unit tests pass** (up from 39 in #299) — 13 new tests:
  - 5 in \`simulated_broker\`: reject rate, partial fill range, stale snapshot, deterministic seeds, EOD tracking
  - 6 in \`replay_report\`: flat / monotonic / drawdown sign / linearized ann_return / empty / TSV roundtrip
  - existing buying-power tests still pass
- [x] End-to-end smoke run with all 3 failure modes verified the divergence and rejection paths actually fire

## Out of scope (follow-ups)

- Golden-snapshot integration test on a fixed date range — needs committed parquet test fixtures or a mock bar source. Unit tests + the failure-injection smoke run cover the same behavioral surface.
- Per-basket-fit metrics in the TSV — the new replay path doesn't expose per-basket P&L.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -1778,6 +1778,7 @@ dependencies = [
  "openquant-core",
  "pair-picker",
  "parquet",
+ "rand 0.8.5",
  "reqwest",
  "serde",
  "serde_json",
--- a/engine/crates/runner/Cargo.toml
+++ b/engine/crates/runner/Cargo.toml
@@ -28,6 +28,7 @@ tokio-tungstenite = { version = "0.24", features = ["native-tls"] }
 futures-util = "0.3"
 arrow = { version = "54", features = ["chrono-tz"] }
 parquet = { version = "54", features = ["arrow"] }
+rand = "0.8"
 
 [dev-dependencies]
 tempfile = "3"
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -436,6 +436,8 @@ pub async fn run_basket_live(
             execution,
         )
         .await?;
+        // Hook for replay's daily-equity time series. Noop on AlpacaClient.
+        broker.record_eod(today).await;
         last_processed_trading_day = Some(today);
         engine.save_state_with_day(state_path, last_processed_trading_day)?;
         processed_sessions.insert(today);
@@ -655,6 +657,9 @@ pub async fn run_basket_live(
                         execution,
                     )
                     .await?;
+                    // Hook for replay's daily-equity time series.
+                    // Noop on AlpacaClient.
+                    broker.record_eod(today).await;
                     last_processed_trading_day = Some(today);
                     engine.save_state_with_day(state_path, last_processed_trading_day)?;
                     info!(
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -1,8 +1,11 @@
-//! Live/paper runner for the basket spread engine.
+//! Live / paper / replay runner for the basket spread engine.
 //!
-//! Drives `basket_engine::BasketEngine` (continuous streaming state machine,
-//! NOT the walk-forward replay in `basket_runner.rs`) with real-time 1-min
-//! bars from the Alpaca WebSocket.
+//! Drives `basket_engine::BasketEngine` (continuous streaming state machine)
+//! with bars from either the Alpaca WebSocket (live, paper) or per-symbol
+//! parquet files via [`crate::parquet_bar_source::ParquetBarSource`] (replay).
+//! All three modes flow through `run_basket_live`; the only difference is
+//! which `Broker`, `BarSource`, `Clock`, and `SessionTrigger` impls are
+//! passed in.
 //!
 //! Flow per trading day:
 //!   1. Startup: load the frozen basket fit artifact and build `BasketEngine`
@@ -503,10 +506,10 @@ pub async fn run_basket_live(
         tokio::select! {
             Some(bar) = bar_rx.recv() => {
                 // `stream.rs` shifts Alpaca bar timestamps by +60s (open→close time).
-                // Undo that here so `minute` reflects bar-OPEN time, matching the
-                // RTH filter used by replay (`basket_runner.rs::read_daily_closes`).
-                // Without this, the last RTH bar (e.g. open=19:59, stream=20:00 in DST)
-                // would be excluded by `RTH_START_MIN..SESSION_CLOSE_MIN` and the close
+                // Undo that here so `minute` reflects bar-OPEN time, matching
+                // `market_session::is_rth_utc` semantics. Without this, the last
+                // RTH bar (e.g. open=19:59, stream=20:00 in DST) would be
+                // excluded by `RTH_START_MIN..SESSION_CLOSE_MIN` and the close
                 // would never enter the buffer — missing the daily close.
                 let bar_open_ts_ms = bar.timestamp - 60_000;
                 let dt = match DateTime::<Utc>::from_timestamp_millis(bar_open_ts_ms) {
--- a/engine/crates/runner/src/broker.rs
+++ b/engine/crates/runner/src/broker.rs
@@ -1,10 +1,11 @@
 //! Broker abstraction for order placement and account queries.
 //!
-//! [`AlpacaClient`] is the production implementation. A `SimulatedBroker`
-//! (follow-up PR) will implement this trait for replay against a synthetic
-//! fill model, letting the live code path run without hitting Alpaca.
+//! [`AlpacaClient`] is the production implementation. The replay-side
+//! `SimulatedBroker` in `simulated_broker.rs` is the other.
 
 use std::collections::HashMap;
+
+use chrono::NaiveDate;
 
 use crate::alpaca::{AlpacaAccount, AlpacaClient, AlpacaOrder, ExecutionMode};
 
@@ -31,6 +32,13 @@ pub trait Broker: Send + Sync {
 
     /// Fetch the account snapshot (status, buying power, equity, gate flags).
     async fn get_account(&self, execution: ExecutionMode) -> Result<AlpacaAccount, String>;
+
+    /// Optional end-of-day equity snapshot hook. Called by `basket_live`
+    /// after `process_session_close` finishes for a trading day. The
+    /// production `AlpacaClient` ignores it (Alpaca already exposes
+    /// historical equity); the replay `SimulatedBroker` records into
+    /// its own time series so the parity TSV writer can read it back.
+    async fn record_eod(&self, _date: NaiveDate) {}
 }
 
 impl Broker for AlpacaClient {
--- a/engine/crates/runner/src/main.rs
+++ b/engine/crates/runner/src/main.rs
@@ -24,6 +24,7 @@ mod clock;
 mod earnings;
 mod market_session;
 mod pair_picker_service;
+mod parity;
 mod parquet_bar_source;
 pub mod refresh;
 mod replay_clock;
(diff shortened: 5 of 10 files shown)


kumargu:
Review result: one material blocker remains.

The replay hardening itself looks directionally right, but the new parity layer is **not actually preserving the legacy quant-lab contract** the PR description claims.

1. `engine/crates/runner/src/parity.rs::compute_stats()` changed the portfolio-stat semantics relative to the deleted `basket_runner.rs`:
- legacy `ann_return` was compounded: `(1.0 + cum).powf(252.0 / n) - 1.0`
- new `ann_return` is linearized: `daily_mean * 252.0`
- legacy `max_drawdown` was stored as a negative return from equity/peak - 1
- new `max_drawdown` is stored as a positive fraction `(peak - equity) / peak`

That means baseline comparison is no longer apples-to-apples even though `BaselineStats` and the PR text present it as the same legacy schema.

2. `parity.rs::write_tsv()` also does **not** match the legacy TSV shape. The old writer emitted per-basket rows first:
`fit_date\tsector\ttarget\tk\tn_trades\tcum_return\tsharpe\tn_days`
and then daily portfolio P&L as:
`date\tdaily_pnl`

The new writer emits only:
`date\tequity\tdaily_pnl_dollar`
with no per-basket section. That may be a reasonable new format, but it is not “TSV format matches legacy writer”, so existing dashboards/harnesses should not be described as preserved unless they are updated too.

I would fix this one of two ways before merge:
- either restore the legacy stats/TSV semantics exactly, or
- explicitly rename/reframe this as a new replay report format and stop claiming baseline/TSV compatibility with the old walk-forward harness.

Without that, the PR is mixing a valid runtime hardening change with a parity-contract regression.

kumargu:
Follow-up: if the intention is **not** to preserve legacy quant-lab compatibility, then my earlier objection should be read as a framing / cleanup issue, not a request to keep the old semantics.

In that case, the cleaner direction is:
- do **not** claim `parity.rs` / TSV / baseline schema are legacy-compatible
- do **not** preserve old output shapes just for historical reasons
- delete any remaining legacy replay/reporting assumptions rather than half-emulating them

Concretely, I would want this PR stack to be explicit that:
- replay metrics are now derived from simulated broker EOD equity, not the old closed-form walk-forward formula
- TSV is a new replay report format, not a drop-in replacement for the deleted `basket_runner` output
- any old “preserve quant-lab parity TSV as-is” wording should be removed if that contract is intentionally being broken

That is a better design than pretending compatibility while drifting semantically.

So from my side the right ask to Claude is: **delete the legacy-compatibility framing and any dead legacy implementation expectations; keep the new replay contract clean and explicit.**

kumargu:
Addressed in d1384e3a:

The parity-layer framing was misleading. Reviewer was right — this isn't legacy-compatible, so claiming compat was wrong. Reframed as a clean new replay report.

**Removed:**
- `parity.rs` renamed to `replay_report.rs`
- `PortfolioStats` → `ReplayStats`
- `BaselineJson`, `BaselineStats`, `print_parity_comparison`, the #256 tolerance check
- `--baseline` CLI flag (replay no longer gates against quant-lab baselines)

**Renamed:**
- `--parity-tsv` → `--report-tsv`

**Doc comments rewritten** to drop "legacy", "matches legacy", "drop-in" claims. State the actual conventions the new report uses (linearized ann_return, positive-fraction max_drawdown).

**New test** `ann_return_is_linearized_not_compounded` documents the chosen semantic so it can't drift silently.

Smoke verified: `replay --engine basket ... --report-tsv ...` writes the report TSV correctly. 52 unit tests still pass; clippy + fmt clean.

The PR description's "matches legacy writer" / "preserves quant-lab parity" wording no longer reflects the code — please re-check the body and let me know if anything else needs reframing.

kumargu:
Re-review after the `replay_report` cleanup:

The legacy-compatibility objection is resolved. Renaming `parity` -> `replay_report`, dropping baseline/legacy claims, and documenting the new conventions is the right direction.

One material issue still remains in the runtime path:

- `run_basket_replay_live_path()` still generates `REPLAY PORTFOLIO STATS` and writes `--report-tsv` **before** checking `if let Err(e) = result`.
- So a replay that fails partway through can still emit or overwrite a report based on an incomplete history.
- That is misleading for operator use and for any downstream comparison scripts, because the command exits non-zero but still leaves a seemingly valid report artifact behind.

I would move the entire report-generation block to after the `result` error check, or guard it explicitly on `result.is_ok()`.

After that, I don’t have another material blocker in this PR.

kumargu:
@codex review

Doc-only cleanup in 93356bdc: removed remaining `basket_runner`/walk-forward/backtest references in module-level doc comments. The replay code now reads as a self-contained new path with no legacy framing. Please confirm there are no remaining material concerns before merge.

kumargu:
Fixed in da582d6d: hoisted the result-error check above the report block. On failure, we log a post-mortem snapshot (still useful for debugging) and exit non-zero — no report TSV written. On success, the report is written as before. So a non-zero exit will never leave behind a seemingly-valid report.

clippy + fmt + 52 unit tests still green.

kumargu (commented):
No remaining material blocker in the code after `da582d6d`.

What changed correctly:
- replay now exits before report generation on failed runs, so a non-zero replay no longer leaves behind a misleading `--report-tsv`
- the `replay_report` rename and doc cleanup make the new contract explicit instead of pretending legacy compatibility
- the replay-side failure injection / EOD equity hooks still look coherent with the live-path reuse design

Residual note only: the PR description is stale and still talks about `parity.rs`, `--parity-tsv`, `--baseline`, and legacy compatibility. That should be updated for readability, but it is not a code blocker from my side.

Issue #306, opened 2026-04-25
[Validation 1] Long-window replay validation (3mo / 6mo / 1yr)


kumargu:
Parent: #305

## Why this exists

We ship replay confident at 1 week (52K bars, 5 sessions). We do not yet know what the new replay path looks like at 3 months / 6 months / 1 year. Things that could go wrong only at scale:

- Memory pressure (parquet pre-load is O(symbols × days × 390 rows))
- Weekend / holiday boundary edge cases
- Reconciliation-error rate drift (divergence is currently rare but at 65× longer windows, even rare events stack)
- Lopsided books (gross getting pinned to one direction over many sessions, which is a known pre-#294 failure mode)
- State-isolation regressions (replay leaking into the live `.state.json` despite #299 protecting against it)

The point of this issue is: run the longer windows, and **document what good looks like** so future runs have a reference.

## Scope

1. Run replay on three windows:
   - 3 months: 2024-07-01 → 2024-09-30
   - 6 months: 2024-04-01 → 2024-09-30
   - 1 year:   2024-01-01 → 2024-12-31
   Each with `--capital 10000`, no failure injection, default knobs otherwise.

2. For each window, capture:
   - `cum_return`, `sharpe`, `max_drawdown`, `daily_mean`, `daily_std`, `n_days`
   - Total accepted / failed orders
   - Number of `BROKER DIVERGENCE` events (should be 0 with no failure injection — any non-zero is a bug)
   - Final positions count
   - Wall-clock runtime (so we can size autoresearch later)

3. Land the report TSVs at `data/replay/baseline_3mo.tsv`, `baseline_6mo.tsv`, `baseline_1yr.tsv`. Add a summary to `docs/REPLAY_BASELINE.md` with the numbers and the exact CLI invocations used.

## Acceptance

- Three TSVs written, each non-empty, each parseable by `replay_report::compute_stats` shape.
- Zero `BROKER DIVERGENCE` errors with default knobs (failure rates 0).
- Equity curve doesn't crash to zero or grow unboundedly.
- `docs/REPLAY_BASELINE.md` checked in with the numbers.

## Out of scope

- Comparing against quant-lab's old baseline numbers — that comparison is invalid because the new path uses different P&L semantics (mark-to-market on simulated fills, not closed-form spread-delta).
- Tuning any knob to improve the numbers. This issue is "what do we get with defaults"; tuning is #2 (autoresearch).
[...]

Issue #305, opened 2026-04-25
Replay validation + autoresearch on the new code path


kumargu:
## Goal

The replay-fix epic (#294) made `replay --engine basket` go through the exact same `run_basket_live()` code path as paper/live. Every line of execution logic is now exercised by replay. This new epic is the *next* step: actually use that to learn things.

Two questions to answer:

1. **Does the new replay path produce sensible numbers on long windows?** We've smoke-tested 1 week and 5 days. We have not run a 3-month or 1-year replay. If anything is structurally wrong (memory pressure, weekend-boundary edge cases, persistent reconciliation divergence at scale, lopsided books) it'll surface here.
2. **What knobs actually matter?** Now that replay is reliable, we can run an autoresearch sweep over the parameter grid — slippage_bps, capital, n_active_baskets, failure-injection rates, basket-fit refresh cadence — and see which ones move P&L meaningfully versus which are noise.

The previous stat-arb autoresearch (2026-04-20, 14 experiments) used the old walk-forward backtest. Findings: baseline Sharpe ≈ 2.80 at 5bps; nothing else helped. That run could not exercise execution-quality knobs because the old path bypassed `process_session_close`. **Re-running autoresearch on the new path is the load-bearing change**: it puts execution quality in the parameter grid for the first time.

## Sub-issues

- **[1] Long-window replay validation** — run replay on 3mo, 6mo, 1yr windows. Verify the equity curve, order count, drawdown, and reconciliation-error rate are all sensible. Land a baseline report TSV in `data/replay/baseline_<window>.tsv` (or a checked-in summary), so we can diff future autoresearch findings against it.
- **[2] Autoresearch sweep on the new replay path** — define the parameter grid (capital ∈ {1k, 10k, 100k}, slippage_bps ∈ {0, 2, 5, 10}, n_active_baskets ∈ {3, 5, 10}, reject_rate ∈ {0, 0.05, 0.10}, partial_fill_rate ∈ {0, 0.10}, basket-fit refresh ∈ {weekly, monthly}). Run the sweep. Record findings: what's a real signal, what's noise.
- **[3] Findings follow-ups** — debug tickets for whatever surfaces. Placeholder.

## Out of scope

- Touching the engine math itself (basket fit, OU/Bertram, signal generation) — that's research-level work, not replay validation.
- Changing the basket universe (`config/basket_universe_v1.toml`) — current universe is the one we're validating against.
[...]

PR #308, opened 2026-04-25
Fix replay leakage: walk-forward fit + drop --fit-artifact / --leverage

kumargu:
Closes part of #306. Refs #305.

## What was wrong

The previous replay path used a single **frozen fit artifact** computed from data through \"now\" (Apr 24, 2026). Replaying any window inside that 60-day fit period produced **fantasy Sharpe numbers** — Q1 2026 came back at **Sharpe 10.66** because ~40% of the replay window was inside the fit's training data. The model already \"knew\" Q1 2026's spreads when it was being asked to trade them.

## Walk-forward fix

\`replay --engine basket\` now **always auto-builds** its fit using data strictly before \`--start\`. The fit physically cannot see the data it's about to trade against.

| Q1 2026 | Sharpe | Cum return | Max DD | Notes |
|---|---|---|---|---|
| Before fix (leaky) | **10.66** | +104% | 2.06% | Fit overlapped replay window |
| After fix (walk-forward) | **3.24** | +27.7% | 6.26% | Fit cut at 2025-12-31 |

3.3× correction. Sharpe 3.24 is now in the same ballpark as quant-lab's full-sample Sharpe 2.80 baseline.

## Changes

- New \`basket_fits::build_replay_fit_artifact_as_of(as_of)\` — builds a fit using data strictly before \`as_of\`.
- New \`basket_live::load_warmup_closes_as_of(as_of)\` helper.
- **Bug fix** in \`load_daily_closes_with_timestamps\`: the lower-bound cutoff was anchored on \`Utc::now()\` while accepting \`max_day_inclusive\` for the upper bound. For replay's as-of cutoff this produced an empty intersection. Now anchors on \`max_day_inclusive\` when provided.
- \`WARMUP_DAYS\` bumped 90 → 180 calendar days. Validator needs ≥ 60 aligned spread points; 90 cal-days ≈ 62 trading days fell below the minimum after alignment dropped any day with a missing symbol. 180 cal-days ≈ 126 trading days is comfortable.
- When \`valid_count == 0\`, now logs per-reason rejection tallies so we see WHY fits failed instead of staring at \"no valid baskets in fit artifact\" blind.

## CLI cleanup

- **Drop \`--fit-artifact\`** from \`replay --engine basket\`. The replay path auto-fits; a stale frozen artifact would leak. Flag still exists on \`live\`/\`paper\` (those expect a curated artifact).
- **Drop \`--leverage\`** from \`replay --engine basket\`. Duplicates \`universe.strategy.leverage_assumed = 4.0\`. Replay must use the same leverage the strategy was designed against — that's config, not a CLI knob.
- \`run_basket_live\`: drop \`fit_artifact_path\` parameter (was only used for one info-log line).

## Q1 2026 walk-forward verification

\`\`\`
cum_return = +27.7%
ann_return = +108%
sharpe     = 3.24
max_dd     = 6.26%
daily_mean = +43 bps
daily_std  = +211 bps
divergence = 0
sessions   = 61
\`\`\`

## Test plan

- [x] All 52 runner unit tests pass
- [x] End-to-end smoke run on Q1 2026 verified Sharpe dropped from 10.66 → 3.24 with fit cut at 2025-12-31

## Follow-ups

- True multi-refit walk-forward (refit every N sessions during replay) — currently the fit is built once at \`--start\` and used for the entire window. For long replays (1yr+) we'll want refit every 60 days. Filing as a separate ticket.
- Re-run Q3 2024 with walk-forward (the prior 0.62 Sharpe used a 21-month-stale frozen fit).

### DIFF
--- a/engine/crates/runner/src/basket_fits.rs
+++ b/engine/crates/runner/src/basket_fits.rs
@@ -10,7 +10,14 @@ use crate::basket_live::{align_basket_history, collect_symbols, load_warmup_clos
 
 const FIT_ARTIFACT_SCHEMA: &str = "basket_fit_artifact";
 const FIT_ARTIFACT_VERSION: &str = "v1";
-const WARMUP_DAYS: i64 = 90;
+// 180 calendar days ≈ 126 trading days. The validator needs at least
+// `residual_window_days = 60` aligned spread points (target + all peers
+// present on the same day). Tight bounds (90 cal days ≈ 62 trading days)
+// fall below that minimum after alignment drops any day with a single
+// missing symbol — which is what made replay's auto-fit reject all 49
+// baskets at first run. Doubling the budget is cheap and removes the
+// edge case entirely.
+const WARMUP_DAYS: i64 = 180;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BasketFitArtifact {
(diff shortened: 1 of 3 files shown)


kumargu (commented):
No material findings on `#308`.

The core fix is the right one:
- replay now builds its fit strictly before `--start`, so it stops leaking in-window data into the model fit
- the lower-bound anchor bug in `load_daily_closes_with_timestamps()` is fixed in the right place
- dropping replay-side `--fit-artifact` is consistent with that contract, because a user-supplied frozen artifact would reintroduce leakage risk
- using `universe.strategy.leverage_assumed` for replay is also coherent if replay is meant to evaluate the strategy as designed rather than a CLI-mutated variant

I also checked the seam with the current runner architecture: this still feeds the same `run_basket_live()` path, so the change is about fit construction, not a forked replay engine.

Residual note only: for long windows, single-fit-from-`--start` is still weaker than a true multi-refit walk-forward, but that is correctly framed as a follow-up, not a blocker in this PR.

PR #317, opened 2026-04-25
Cap entries at engine, size off live equity, fix pnl_per_unit sign

kumargu:
## Summary

Three related fixes the Q4 2024/2025 replays surfaced.

### 1. Active-basket cap moved to engine entry-time (FCFS admission)

Previously every session admitted all 43 baskets and the portfolio layer flattened the lowest-|z| ones. But mean-reverting positions have shrinking |z| as they pay off, so the portfolio cap kicked them out exactly when they were about to be profitable.

Q4 2025 telemetry pre-fix: 2,153 transitions, **0 flips** (every transition was \`initial_entry_*\`). The cap-driven churn meant baskets never lived long enough to mean-revert.

With the cap at entry (\`BasketEngine::set_max_active_positions\` + \`active_position_count\` snapshot inside \`on_bars\`), positions live until the engine's own Bertram exit (a flip at the opposite z=±k) fires. Q4 2025 post-fix: \`active_after=5\` stable, real flips visible (\`flip_long_to_short\`, \`flip_short_to_long\`).

### 2. Dynamic per-basket notional from live equity

\`plan_portfolio_for_equity\` sizes notional off the broker's current equity at session-close, with a \`BUYING_POWER_UTILIZATION = 0.90\` buffer.

Static \`config.capital\` ignored drawdowns: a 4% loss left target gross above \`equity × leverage\`, broker rejected orders piecemeal, the hedge book ended up lopsided, next day's loss widened the gap. Q4 2024 was the loudest example. The 90% buffer absorbs per-symbol rounding + plan/submit price drift so the actual submitted gross stays under buying power.

### 3. \`pnl_per_unit_notional\` sign-flip in transition log

Spread is \`log(target) - mean(log(peers))\`. A LONG basket holds target long and peers short via \`basket_to_legs\`, so it profits when target outperforms peers — i.e. when spread **rises**. The old formula negated \`old_pos\` and reported every gain as a loss. Telemetry-only, no impact on actual fills, but it had me chasing a phantom bug.

### Plus

- Clippy fix in \`pair-picker/src/scorer.rs\` (\`manual_range_contains\`) so the workspace builds clean under Rust 1.94.

## Q4 2025 replay (cap-fix only baseline)

| metric | value |
|---|---|
| cumulative return | -2.49% |
| annualized return | -8.79% |
| Sharpe | -0.54 |
| Max drawdown | -10.20% |
| trading days | 64 |

This is the **honest baseline** — the cap-fix surfaces a real strategy issue this PR doesn't address: 4 of 5 active baskets get stuck far past their flip thresholds (PNC long entered z=-1.12 drifted to z=-3.91, TFC short entered z=+0.49 drifted to z=+4.18) because Bertram symmetric has no exit when cointegration breaks. Follow-up PR adds adverse-move stop-loss + re-arm gate (lab-validated 2.0σ default).

## Test plan

- [x] \`cargo test -p basket-engine\` — 23 lib tests pass (added \`test_notional_scales_with_dynamic_equity\`, \`test_plan_portfolio_enforces_active_basket_cap\`, \`test_portfolio_config_rejects_zero_active_baskets\`)
- [x] \`cargo test --workspace\` passes
- [x] \`cargo build --release -p openquant-runner\` clean
- [x] Q4 2025 replay end-to-end clean (TSV: \`data/replay/q4_2025_capfix.tsv\`)

### DIFF
--- a/config/basket_universe_v1_no_mining.toml
+++ b/config/basket_universe_v1_no_mining.toml
@@ -1,0 +1,106 @@
+# basket_universe_v1_no_mining.toml — RESEARCH VARIANT (NOT FROZEN)
+# ----------------------------------------------------------------------
+# This is `basket_universe_v1.toml` with the `sectors.mining` block
+# removed. It exists for replay-only research while the metals/mining
+# basket is being redesigned.
+#
+# Why mining was pulled (not deleted from v1, just excluded here):
+#   `sectors.mining` mixed three macro drivers:
+#     - NEM, FCX  — gold / copper miners (commodity price-driven)
+#     - NUE, STLD — steel producers (industrial demand-driven)
+#     - MLM, VMC  — construction aggregates (infrastructure-driven)
+#   Cointegration is fragile across these regimes; we'd rather not
+#   trade a basket whose peers decouple by design.
+#
+# Status:
+#   - Replay only. NOT for live/paper.
+#   - Frozen v1 is untouched. The autoresearch baseline (Sharpe 2.80,
+#     point estimate from full v1) still applies to v1, NOT to this.
+#   - Use this via `replay --engine basket --universe config/basket_universe_v1_no_mining.toml`.
+#   - Quant-lab is researching a proper metals replacement
+#     (precious-only / industrial-only / aggregates-only sub-baskets).
+#     When that lands, a frozen v2 supersedes both this file and v1.
+#
+# DO NOT promote this file to live until quant-lab has signed off and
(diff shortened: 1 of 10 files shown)


kumargu:
Fixed in 294b388b. The gross-cap guard now uses `sizing_equity * leverage` (matching the dynamic-sizing path) and the error message includes live equity for diagnostics. Caught my own miss — thanks for the precise call-out.

Issue #321, opened 2026-05-02
Autoresearch: test one-shot basket entry timing and close-snapshot robustness


kumargu:
## Summary

Current basket live/paper behavior is effectively a one-shot daily decision:

- collect 1-min bars through the session
- keep the latest valid RTH close per symbol
- at session close + grace, form one daily snapshot
- compute spread / z-score from that snapshot
- enter / hold / flip once

That is coherent as a daily strategy, but it leaves an open research question:

**If we want to trade once per day to stay less noisy, what is the right time and the right end-of-day price snapshot for that one decision?**

The current implementation answers only one version of that question:

- **decision time** = session close + grace
- **decision price** = last 1-min bar for each symbol

We should test whether that is actually the best daily timing, or whether a nearby pre-close snapshot / more robust close estimator works better.

## Motivation

Recent paper/paper-recovery review raised two concerns:

1. **Last-bar dependence may be too noisy**
   - A daily decision can be sensible.
   - But using only the final 1-min close may overreact to closing noise / late prints / microstructure effects.

2. **We may want one decision per day, but not necessarily at the official close**
   - `t-15m` or `t-5m` may be more stable than `close + grace`.
   - A short VWAP / median-of-last-bars style snapshot may be more robust than the single final bar.

This issue is specifically about **timing and snapshot robustness**, not about moving to full intraday trading.

## Hypothesis

A better daily basket strategy may be:

- still **one decision per day**
- but with a better **decision time** and/or **decision snapshot** than `last 1-min bar at close + grace`

My prior is:

- `close + grace` may be too dependent on the final print
- `t-5m` or `t-15m` may be more stable
- `last-5m VWAP` or `last-15m VWAP` is a plausible winner versus `last bar only`
- full intraday exit/entry research should come later, after this simpler question is answered

### Phase 1: one-shot decision time sweep

Keep everything else fixed. Only vary the **time of the once-per-day decision**.

Variants:

- `t-30m`
- `t-15m`
- `t-5m`
- `close`
- `close + 2m grace` (current behavior)

Interpretation:

- This isolates timing only.
- If performance changes materially, the current close timing is not robust.

### Phase 2: end-of-day snapshot robustness

Keep one daily decision, but change the price estimator used near the close.

Variants:

- `last 1-min bar` (baseline)
- `last 5m VWAP`
- `last 15m VWAP`
- `median of last 5 1-min closes`
[...]

PR #322, opened 2026-05-02
Fix replay session-close race (SharedCloses leak)

kumargu:
## Summary

Replay had a session-close race that produced ~$15 cash drift between identical runs, varying with tokio scheduling. Smoke matrix on FAANG / 2025-07-01..2025-07-03 (3-day window):

| matrix | distinct outputs | example drift |
|---|---|---|
| pre-fix sequential ×3 | 2 | Sharpe -6.55 vs -10.70 |
| pre-fix parallel ×3 | 2 | cash $14.49 |
| **post-fix all 6** | **1** | byte-identical (md5 ×6) |

## Cause

`SimulatedBroker.fill_price()` reads `SharedCloses` at fill time, but the bar emitter task keeps writing to `SharedCloses` while the consumer awaits inside `process_session_close` (seed_current_shares, place_order, the 30s post-submit reconciliation sleep). Each `.await` is a yield point where the emitter advances and overwrites `SharedCloses` with bars from the *next* day. Fills then land at prices the engine never saw.

The simulated_broker docstring already claimed the intended contract — "Fills happen at the latest known close … the close from the bar that triggered the session-close cycle" — but the implementation didn't enforce it across yield points.

## Fix

Add a back-channel `done_rx` from the consumer to the bar emitter:

- `SessionTrigger::ack_session_processed(&mut self)` — new trait method, default no-op for live's `IntervalSessionTrigger`. `BarDrivenSessionTrigger` overrides to send `()` on a cap=1 mpsc channel.
- `parquet_bar_source::emit_loop` calls the renamed `drain_signal_and_wait_ack` at every date-cross AND at replay exhaustion. The emitter now blocks on `done_rx.recv()` after each `session_tx.send(date)`, so `SharedCloses` stays frozen at D's last close until the consumer signals "fills + record_eod + save_state are all done."
- `basket_live`'s session_event arm calls `session_trigger.ack_session_processed().await` after `engine.save_state_with_day(...)` succeeds. Failure paths (the `?` operators) skip the ack on purpose: the runner exits, `done_tx` drops, the emitter sees `None` from `done_rx.recv()` and unwinds cleanly.

## Live impact

None. The default `ack_session_processed` is a no-op, and only `BarDrivenSessionTrigger` (replay-only) overrides it. Live/paper uses `IntervalSessionTrigger` on the wall-clock path — no behavior change.

## Test plan

- [x] `replay_clock::tests::ack_drives_done_channel` — pins the ack→done_rx wiring so a future refactor can't silently regress the race.
- [x] `replay_clock::tests::ack_does_not_panic_when_emitter_gone` — consumer-survives-emitter-exit invariant.
- [x] All existing 55 runner unit tests still pass.
- [x] 6-run smoke matrix (3 sequential + 3 parallel) on FAANG / Q3-start-3-day window: all byte-identical.
- [ ] CI: cargo fmt --check, cargo clippy -D warnings, cargo test --workspace.

## Why this matters

Found while investigating issue #321 (decision-time / snapshot-robustness experiment). Without this fix, the experiment's 4-variant Sharpe comparison is dominated by ~3-4 Sharpe-points of replay noise, swamping any real offset effect. With it, replay is bit-exact reproducible — sequential and parallel.

This bug also retroactively adds a noise floor to **every existing autoresearch result** that ran through this code path (the entire monetization-matrix sweep, λ-sweeps, etc.). The signal in those experiments is real, but quoted Sharpes have ≥0.5 of noise we didn't know about until now.

### DIFF
--- a/engine/crates/runner/src/basket_journal.rs
+++ b/engine/crates/runner/src/basket_journal.rs
@@ -190,7 +190,10 @@ impl BasketJournal {
     }
 
     pub fn record_run(&self, rec: &BasketRunRecord<'_>) -> Result<(), String> {
-        let conn = self.conn.lock().map_err(|_| "basket journal mutex poisoned".to_string())?;
+        let conn = self
+            .conn
+            .lock()
+            .map_err(|_| "basket journal mutex poisoned".to_string())?;
         conn.execute(
             "INSERT OR REPLACE INTO basket_runs (
                 run_id, started_at_utc, execution_mode, universe_path, fit_artifact_path,
@@ -219,7 +222,10 @@ impl BasketJournal {
     }
 
     pub fn record_session_close(&self, rec: &BasketSessionCloseRecord<'_>) -> Result<(), String> {
-        let conn = self.conn.lock().map_err(|_| "basket journal mutex poisoned".to_string())?;
+        let conn = self
+            .conn
+            .lock()
+            .map_err(|_| "basket journal mutex poisoned".to_string())?;
         conn.execute(
             "INSERT INTO basket_session_closes (
                 run_id, trading_day, status, closes_received, symbols_expected,
@@ -289,7 +295,10 @@ impl BasketJournal {
     }
 
     pub fn record_order_event(&self, rec: &BasketOrderEvent<'_>) -> Result<(), String> {
-        let conn = self.conn.lock().map_err(|_| "basket journal mutex poisoned".to_string())?;
+        let conn = self
+            .conn
+            .lock()
+            .map_err(|_| "basket journal mutex poisoned".to_string())?;
         conn.execute(
             "INSERT INTO basket_order_events (
                 run_id, trading_day, seq, symbol, side, requested_qty, intended_notional,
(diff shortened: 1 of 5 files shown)


kumargu:
Addressed in f87d34b8:

**Codex P1 (legit bug):** Restructured the `session_event` arm so the ack is reached on every consumed event (including the dedup path and the non-trading-day path that previously `continue`-ed past the ack). Verified end-to-end: ran with `--start 2025-07-01 --end 2025-07-01` to leave state at `last_processed=2025-07-01`, then re-ran with `--start 2025-07-01 --end 2025-07-02 --resume-state` — hits the new `session-close event for already-processed date — acknowledging without rerun` log path and exits cleanly in 32s. Pre-fix this hangs forever on the missing ack.

**Race-fix determinism re-verified after restructure:** 1 sequential + 2 parallel runs all produce md5 `fb56ff0d2ebcfd09902fc2ea8f814830`.

**CI fmt failure** was pre-existing on `origin/main` (drift in `basket_journal.rs` from 0ca79878 + an import-block reshape in `basket_live.rs`); ran `cargo fmt --all` — formatter-only diffs, no semantics.

PR #323, opened 2026-05-03
issue #321: walk-forward frozen-rule research + dom060 strategy candidate

kumargu:
Closes the research portion of #321. Two-track scope per [reviewer plan](https://github.com/kumargu/OpenQuant/issues/321#issuecomment-4366774127):

1. **Strategy candidate** — `dom060` (`config/basket_universe_v1_no_mining_statgate2_dom_only.toml`), validated under frozen-rule walk-forward across 4 OOS blocks (2 of which are truly out-of-sample).
2. **Research / reproducibility tooling** — analysis scripts, `freeze-basket-fits --as-of` flag, walk-forward output artifacts, bar audit.

Resume tag for this state: `wf-frozen-rule-baseline-v1` (commit `1af64ce7`).

## Walk-forward results (Sharpe, $10k cap, 8 active baskets, 5 bps cost)

| variant | test0 (truly OOS) | test1 (truly OOS) | test2 | test3 | mean | min |
|---|---|---|---|---|---|---|
| baseline | +1.21 | +0.49 | +0.10 | +3.33 | +1.28 | +0.10 |
| dom050 | +1.18 | +2.86 | +2.80 | +1.63 | +2.12 | +1.18 |
| **dom060** | **+1.79** | **+3.08** | +1.64 | +2.73 | **+2.31** | **+1.64** |
| nomegacaps | +2.26 | +1.10 | +0.21 | +4.64 | +2.05 | +0.21 |
| stack (dom050+nomegacaps) | +1.99 | +2.87 | +2.84 | +2.04 | +2.43 | +1.99 |

| block | window | provenance |
|---|---|---|
| test0 | 2024-07-01 → 2024-12-31 | **truly OOS** for both dom thresholds |
| test1 | 2025-01-01 → 2025-06-30 | **truly OOS** for both |
| test2 | 2025-07-01 → 2025-12-31 | **in-sample for both** dom050 and dom060 (the q3_q1 sweep that produced both thresholds spans this block) |
| test3 | 2026-01-01 → 2026-03-31 | **in-sample for both** dom050 and dom060 (same) |

**Decision rule pass:** dom060 lifts baseline by **+0.58 / +2.59** on the two truly-OOS blocks (test0 and test1). dom050 fails Test0 (lift -0.03). dom060 chosen over stack — both clear the bar but dom060 keeps 10–15 valid baskets per block vs stack's 4–9.

Quarterly Sharpe for dom060: **+1.24 / +2.35 / +3.39 / +2.77 / +1.04 / +2.26 / +2.73** (Q3 2024 → Q1 2026, all positive). Reproducible from `autoresearch/issue_321/quarterly_sharpe.py`.

### Strategy candidate
- `config/basket_universe_v1_no_mining_statgate2_dom_only.toml` — the validated `dom060` candidate (drops baskets where any single member owns >60 % of spread variance)
- Live engine defaults unchanged. The existing `dominance_gate_enabled` / `dominance_max` knobs (added in `ca3826e9`) are off by default; this config opts in.

### Research / reproducibility (per the reviewer's checklist)
- `engine/crates/runner/src/main.rs` — `freeze-basket-fits --as-of YYYY-MM-DD` flag so per-block reject lists can be reproduced after the fact (replay never persists its in-memory walk-forward fit). Future-dated `--as-of` now errors immediately rather than continuing to "zero valid baskets".
- `autoresearch/issue_321/analyze_walkforward.py` — the flat per-(variant, block) table + Jaccard reject-list stability + dom×nomegacaps additivity check + binary OOS-lift verdict. **`orders` column corrected** to count `ORDER PLACED` lines (real broker emissions); a separate `intents` column counts `BASKET_INTENT` (decisions/intents).
- `autoresearch/issue_321/audit_bars.py` — bar-completeness audit. Verifies decision-time close-bar (19:55–20:00 UTC) coverage **for all 4 blocks** and all 6 mega-caps. Coverage is **96.9 %** in test0 and **≥97.7 %** in test1–test3.
- `autoresearch/issue_321/build_per_block_fits.sh` — driver that calls `freeze-basket-fits --as-of` for every (variant, block) so the per-block fit JSONs match what each replay computed.
- `autoresearch/issue_321/quarterly_sharpe.py` — quarter-by-quarter Sharpe per variant from the daily P&L in each `report.tsv`.
- `autoresearch/issue_321/walkforward/<variant>/<block>/` — replay outputs (engine.log, report.tsv, replay state) for all 5 variants × 4 blocks = 20 runs.
- `autoresearch/issue_321/walkforward_fits/<variant>/<block>_fit.json` — frozen per-block fit artifacts.
- `autoresearch/issue_321/walkforward_full_report.txt` — full analyzer dump.
- Companion research configs: `basket_universe_v1_no_mining_baseline.toml`, `…_dom050.toml`, `…_dom055.toml`, `…_dom065.toml`, `…_dom070.toml`, `…_dom080.toml`, `…_no_megacaps.toml`, `…_no_megacaps_dom050.toml`, `…_random9_seed{0..4}.toml`.

### Self-review fixes (commit `9bb3dbfb`)
- `cargo fmt` and `cargo clippy --workspace -- -D warnings` now pass locally.
- ADF critical-value table swap: the metric was using `EG_CRITICAL_VALUES` (Engle-Granger 2-step, k=1, designed for OLS-residuals). Our spread is fixed-weight (no OLS), so the correct table is standard ADF with constant (MacKinnon 1996): 1%=-3.435, 5%=-2.864, 10%=-2.568. Renamed to `ADF_CONSTANT_CRITICAL_VALUES` with citation. Old table was conservative (over-rejects), so dom060 OOS Sharpe numbers are not affected; only stored `adf_pvalue` fields move.
- 5 new validator tests cover the ADF and dominance gates in their enabled state (admit-when-passing and reject-when-failing paths, plus the `score unavailable` rejection).
- `DontDoConfig` removed (was parsed from TOML but never read in code; `[dont_do]` blocks in TOMLs are now silently ignored).

### Older history riding along (unmerged on main)
- `66269f8c` — `[runner].decision_offset_minutes_before_close` TOML knob
- `75bfec21` — replay determinism (HashMap → BTreeMap in float-sum hot paths). Same fix as separate PR #316; coordinate merge order.
- `cd2fcf8f`, `dd4f4fcf`, `6f334bed` — earlier #321 research checkpoints (breakout-count analysis + research-team memos). These are research artifacts, not code paths.
- `ca3826e9` — Codex's ADF + dominance admission diagnostics (the gate plumbing this PR's strategy candidate uses).

## Reproducing locally

```bash
git checkout wf-frozen-rule-baseline-v1   # or the head of this PR
cargo build --release -p openquant-runner
./autoresearch/issue_321/build_per_block_fits.sh        # generates per-block fits
python3 autoresearch/issue_321/analyze_walkforward.py   # full table + verdict
python3 autoresearch/issue_321/quarterly_sharpe.py      # quarterly numbers
python3 autoresearch/issue_321/audit_bars.py            # bar-completeness audit
```

## Test plan

- [x] CI green (`cargo fmt --check`, `cargo clippy -D warnings`, `cargo test --workspace`)
- [x] Reviewer check: orders column is real broker orders, intents column is `BASKET_INTENT` count
- [x] Reviewer check: bar audit shows close-bar coverage ≥96.9% at decision time across all 4 blocks
- [x] Reviewer check: Jaccard stability of dom060's gate-reject list (0.50–0.65 across the 4 blocks; 12 always-rejected names)
- [ ] Sign off on `dom060` (vs `stack`) as the validated candidate — see open question on min-Sharpe vs diversification


kumargu:
## P0 — CI is failing; do not merge

`gh pr checks 323` shows **Rust tests + lint = pending/failed**. Two separate failures confirmed locally:

### 1. `cargo fmt --check` fails

Two files have formatting violations:

**`engine/crates/basket-picker/src/validator.rs` line 136** — single-line `None =>` arm in a match block needs to be wrapped in braces per rustfmt:
```
-            None => return BasketFit::rejected(candidate.clone(), "ADF gate failed: test unavailable"),
+            None => {
+                return BasketFit::rejected(candidate.clone(), "ADF gate failed: test unavailable")
+            }
```

**`engine/crates/runner/src/basket_fits.rs` line 86** — the `ValidatorConfig { ... }` struct literal inside `build_live_fit_artifact_from_inputs` is indented with an extra 4-space offset (looks like it was inside a block that got removed). rustfmt expects function-body indentation, not block-body indentation.

### 2. `cargo clippy -D warnings` fails

`engine/crates/basket-picker/src/dominance.rs:43` — `needless_range_loop` on the covariance loop:
```rust
for t in 0..n {
    s += (cols[i][t] - means[i]) * (cols[j][t] - means[j]);
}
```
Clippy wants an iterator form. The existing `adf.rs` already has `#[allow(clippy::needless_range_loop)]` for the identical pattern — `dominance.rs` just needs the same suppress attribute on the outer covariance loop, or the loop needs to be rewritten to satisfy clippy.

Fix both and re-push before any further review.

kumargu:
## P1 — OOS classification is wrong for dom060 in the provenance table

The PR description's block provenance table says:

| block | window | provenance |
|---|---|---|
| test2 | 2025-07-01 → 2025-12-31 | in-sample (dom050 was tuned on a window starting here) |
| test3 | 2026-01-01 → 2026-03-31 | in-sample (same) |

The parenthetical "dom050 was tuned" implies this applies only to dom050. **That's wrong. dom060's threshold (0.60) was also selected from a sweep on the 2025-07-01 → 2026-03-31 window** (the `q3_q1_statgate2/dom_only/` run that produced Sharpe 1.568, as documented in the issue comments). Both dom050 and dom060 are in-sample on test2 and test3.

**What this changes:**
- The truly-OOS results for dom060 are test0 (+1.79) and test1 (+3.08) only — these match the claimed +0.58 / +2.59 lifts and are genuinely OOS. That part is fine.
- test2 (+1.638) and test3 (+2.731) are also in-sample for dom060. The PR's min Sharpe of +1.64 and the mean of +2.31 reflect in-sample performance too. This doesn't invalidate the strategy — the test0/test1 OOS lift is the key gate — but the framing in the description is misleading.

**Required fix:** Update the provenance table to say "in-sample (both dom050 and dom060 thresholds were selected from a sweep on this window)." The OOS lift claim (+0.58 / +2.59 on test0 + test1) is still valid and stands.

kumargu:
## P1 — Wrong ADF critical value table

**File:** `engine/crates/basket-picker/src/adf.rs`

```rust
const EG_CRITICAL_VALUES: [(f64, f64); 4] =
    [(0.01, -3.90), (0.05, -3.34), (0.10, -3.04), (0.25, -2.58)];
```

These are **Engle-Granger cointegration residual test** critical values (MacKinnon 1994, k=1 cointegrating variable). The EG test is a two-step procedure: (1) estimate the cointegrating vector by OLS, then (2) run ADF on the OLS residuals using these more-negative critical values to account for the estimated parameters.

The basket spread is computed with a **fixed equal-weight formula** (`log(target) − mean(log(peers))`), not by OLS estimation. This means the EG critical values are inappropriate here. The correct table is the standard ADF with constant (MacKinnon 1994, k=0): 1%=−3.435, 5%=−2.864, 10%=−2.568.

**Direction of bias:** The EG values are ~0.5 units more negative, so the gate will **reject too many baskets** (false-negatives). For an admission gate this is conservative, not catastrophically wrong — but the p-values it reports are overstated. A basket with ADF stat of −3.00 would show p≈0.20 under EG critical values but p≈0.05 under standard ADF critical values. Anyone using the stored `adf_pvalue` in the fit JSON for downstream analysis will get wrong numbers.

**Required fix:** Either:
1. Replace `EG_CRITICAL_VALUES` with standard ADF critical values for a model with constant (no trend), rename the constant appropriately, and add a comment citing MacKinnon 1994 Table 1 row for n→∞ without trend.
2. Or keep EG values but add a comment documenting the conservative choice and why it's intentional, and acknowledge the p-value overstating in the schema docs for `adf_pvalue`.

kumargu:
## P1 — New admission gate code has no tests in the enabled state

**Files:** `engine/crates/basket-picker/src/validator.rs`, `engine/crates/basket-picker/src/adf.rs`, `engine/crates/basket-picker/src/dominance.rs`

All four existing tests in `validator.rs` use `ValidatorConfig::default()`, which has both gates disabled (`adf_gate_enabled = false`, `dominance_gate_enabled = false`). The ADF and dominance gate code paths are dead in every test.

Missing coverage:

1. **`adf_gate` rejects a non-stationary series.** The `adf.rs` module has one test for a stationary series that PASSES. There's no test that a random-walk (unit root) series is correctly REJECTED when `adf_gate_enabled = true`.

2. **`adf_gate` rejection path in `validator.rs`.** No test calls `validate()` with `adf_gate_enabled = true` and a non-stationary spread. Should verify: `fit.valid == false` and `fit.reject_reason.contains("ADF gate failed")`.

3. **`dominance_gate` rejection path.** The `dominance.rs` module has one test where the target dominates — but no test calls `validate()` with `dominance_gate_enabled = true` and an over-concentrated basket. Should verify the rejection reason matches.

4. **`dominance_gate None` path.** `max_component_dominance` returns `None` for n<10 or mismatched lengths. When the gate is enabled and the score is unavailable, the basket is rejected with "score unavailable". This path has no test.

5. **`build_replay_fit_artifact_as_of`.** The new `--as-of` path in `basket_fits.rs` has no unit test. A test could pass a synthetic `bars_dir` (via temp dir with minimal parquets) and verify the returned artifact has `generated_at` in the past relative to `as_of`.

The "no tech debt" standard requires covering the rejection paths. With this code going into production as the dom060 admission filter, silent bugs in the gate logic lose real money.

kumargu:
## P2 — dom060 vs stack decision: diversification argument is not quantified

The PR recommends dom060 over stack because dom060 "keeps 10–15 valid baskets per block vs stack's 4–9 (concentration risk in test2)." But the numbers tell a different story:

| block | dom060 n_valid | dom060 Sharpe | stack (nomegacaps+dom050) n_valid | stack Sharpe |
|---|---|---|---|---|
| test0 | 15 | +1.79 | 7 | +1.99 |
| test1 | 12 | +3.08 | 9 | +2.87 |
| test2 | 10 | +1.64 | 4 | **+2.84** |
| test3 | 11 | +2.73 | 9 | +2.04 |
| **min** | | **+1.64** | | **+1.99** |

In test2 — the block cited as evidence of concentration risk — stack had only 4 valid baskets and delivered Sharpe **2.84**, while dom060 had 10 valid baskets and delivered Sharpe **1.64**. Concentration in that window wasn't a problem; it was better. The stack's min Sharpe (+1.99) is also higher than dom060's min (+1.64) by 0.35 across all four blocks.

The "diversification" argument amounts to: "in an unseen future block where all 4 valid stack baskets simultaneously fail, we'd prefer dom060." That's forward-looking speculation, not evidence from this data.

**The PR should either:**
1. Choose stack and acknowledge the min-Sharpe dominance, OR
2. Keep dom060 but make an explicit argument: e.g., "stack's test2 4-basket book is below our minimum operational threshold for a well-diversified position" and name that threshold (3 baskets? 5?), OR
3. Acknowledge this is a judgment call and let the reviewer make it explicitly, rather than presenting the diversification argument as if it's data-driven.

Not a blocker, but the claim as written is hand-waving.

kumargu:
## P2 — DontDoConfig is dead code introduced in this PR

**File:** `engine/crates/basket-picker/src/universe.rs`

Commit `66269f8c` (in this PR branch) introduced `DontDoConfig`:

```rust
pub struct DontDoConfig {
    pub hl_trade_gate: bool,
    pub time_based_exit: bool,
    pub stop_loss: bool,
    ...
}
```

It is parsed from TOML (the `[dont_do]` section), stored on `Universe`, and never read anywhere. `grep -rn 'dont_do\.' engine/crates/` returns zero results outside of the definition/constructor.

Per the "no dead code" standard: if DontDoConfig is documentation-only, it should either be a comment block in the TOML file (not Rust code that compiles with warnings silently), or its field values should be enforced in the validator or engine. Parsing 8 booleans into a struct and then ignoring them wastes reader attention and creates confusion about whether the engine actually enforces these constraints.

Either wire at least one field up (e.g., validate the config and return an error if a prohibited flag combination is present), or remove the struct and leave the TOML comments as plain TOML comments.

kumargu:
## Minor findings — batch

**1. `build_per_block_fits.sh` stale count**

The script ends with `echo "all 15 per-block fits built"` but iterates 5 variants × 4 blocks = **20** builds. The "15" was left over from when the loop only had 3 blocks (before Test0 was added). Fix the string.

**2. `audit_bars.py` does not cover Test0**

```python
BLOCKS = [
    ("Test1", "2025-01-01", "2025-06-30"),
    ("Test2", "2025-07-01", "2025-12-31"),
    ("Test3", "2026-01-01", "2026-03-31"),
]
```

Test0 (2024-07-01 → 2024-12-31) is missing. The PR claims ≥98.8% bar coverage but that figure was derived from Test1-3. Test0 uses H2 2024 data from a different period and the parquet quality there is unaudited. Add Test0 to the block list.

Additionally, the audit counts full-day bars <380 (out of 390 expected) but the strategy makes decisions at session close (last bar ~19:59 UTC). A day could have 385 bars and still be missing the close bar. The audit should also explicitly check for the presence of at least one bar in the 19:55–20:00 UTC window per day.

**3. PR #316 (determinism fix) still open — same files**

Commit `75bfec21` in this branch is the same fix as PR #316 (`fix/replay-determinism`). Both touch `engine/crates/basket-engine/src/engine.rs` and `engine/crates/runner/src/simulated_broker.rs`. If #316 merges first, this branch will need a rebase to avoid a merge conflict. Consider coordinating: if this PR lands first, #316 should be closed; if #316 lands first, squash out the duplicate commit from this branch.

**4. `--as-of` with future date fails with unhelpful error**

`freeze-basket-fits --as-of 2030-01-01` will load zero close prices (no parquets exist for 2030), trigger the "refusing to write basket fit artifact with zero valid baskets" guard, and exit(1). The error message doesn't tell the user why there are zero valid baskets. A check like `if as_of > Utc::now().date_naive() { error!(..., "--as-of date is in the future"); ... }` would save confusion.

**5. 7 quarterly Sharpe numbers are not reproducible from stored artifacts**

The PR description states: "Quarterly Sharpe for dom060: +1.24 / +2.35 / +3.39 / +2.77 / +1.04 / +2.26 / +2.73 (Q3 2024 → Q1 2026, all positive)." The 4-block reports can produce 2 half-yearly and 2 more blocks, not 7 separate quarterly figures. No script in `autoresearch/issue_321/` computes per-quarter Sharpe from the daily equity curves. These numbers cannot be reproduced by anyone checking out the tag. Either add a script that derives them from the stored `report.tsv` files, or remove the claim.

kumargu:
## Summary review — severity graded

### P0 — Must fix before merge (CI gate, not merging with failing CI)

1. **`cargo fmt --check` fails** — two files: `validator.rs:136` (match arm braces) and `basket_fits.rs:86` (indentation). Run `cargo fmt --all` and push.

2. **`cargo clippy -D warnings` fails** — `dominance.rs:43` `needless_range_loop`. Add `#[allow(clippy::needless_range_loop)]` (matching the existing suppress in `adf.rs`) or rewrite as iterator.

---

### P1 — Must address before merge (correctness, methodology, test coverage)

3. **OOS classification is misleading** — the provenance table says "in-sample (dom050 was tuned on a window starting here)" for test2/test3, but dom060's threshold (0.60) was also selected from a sweep on that window (`q3_q1_statgate2/dom_only/` run at Jul'25–Mar'26 that produced Sharpe 1.568). The OOS lift claim (+0.58 / +2.59 on test0 + test1) is valid and stands, but the provenance table must say both dom050 and dom060 are in-sample on test2/test3.

4. **Wrong ADF critical value table** — `EG_CRITICAL_VALUES` (MacKinnon 1994 EG cointegration test, k=1) is the two-step Engle-Granger table designed for OLS regression residuals. The basket spread uses a fixed equal-weight formula, not OLS estimation. The correct table is standard ADF with constant (k=0): 1%=−3.435, 5%=−2.864, 10%=−2.568. The bias is conservative (too many rejections), not permissive, but stored `adf_pvalue` fields are systematically overstated by ~0.5 critical value units. Either fix the table (preferred) or document the conservatism explicitly.

5. **No tests for adf_gate or dominance_gate in enabled state** — all four `validator.rs` tests use default config (both gates off). Missing: a test that a non-stationary spread is rejected when `adf_gate_enabled=true`; a test that a dominant-target basket is rejected when `dominance_gate_enabled=true`; a test for the `dominance score unavailable` rejection path.

---

### P2 — Should address before merge (methodology, dead code)

6. **dom060 vs stack recommendation is hand-waving** — stack has higher min Sharpe across all 4 blocks (+1.99 vs +1.64), and in the cited test2 "concentration" block the 4-basket stack actually outperformed dom060 (2.84 vs 1.64). The diversification argument needs quantification or the recommendation should be revised.

7. **`DontDoConfig` is dead code** — parsed from TOML, stored on `Universe`, never read anywhere in the engine or validator. Either wire at least one field or remove the struct.

---

### Minor — Nice to have

8. `build_per_block_fits.sh` says "all 15 per-block fits built" — actually 20 (5 variants × 4 blocks). Stale string from before Test0 was added.

9. `audit_bars.py` covers Test1-3 only — Test0 (H2 2024) unaudited. Also, the audit counts full-day bars <380 but doesn't verify the close bar at 19:55–20:00 UTC specifically exists.

10. PR #316 (replay determinism fix) is still open and modifies the same two files. Coordinate to avoid a merge conflict.

11. `--as-of` accepts future dates, fails with zero-valid-baskets error instead of an actionable message.

12. 7 quarterly Sharpe numbers in PR description are not reproducible from stored artifacts.

---

### What's confirmed correct (don't manufacture findings)

- The Sharpe numbers in the PR table exactly match the `report.tsv` headers (`sharpe=1.787568` rounds to +1.79, etc.).
- The `orders` column fix is verified: `grep -c "ORDER PLACED" baseline/test1/journal/engine.log = 1315` matches the table exactly. Same for dom050/test1 (orders=1498, intents=291).
- The OOS lift claim (+0.58 / +2.59 on the two truly-OOS blocks) is arithmetically correct.
- The always-rejected (12) names for dom060 are correctly computed by the analyzer.
- The date parsing for `--as-of` uses `chrono::NaiveDate::parse_from_str` with "%Y-%m-%d" — safe, no silent truncation.
- The `max_component_dominance` function's NaN guard (`!total_var.is_finite()`) correctly handles the zero-price edge case (returns None rather than wrong score).
- The interpolation direction in `interpolate_p_value` is correct (more negative stat → smaller p).

The core walk-forward methodology (frozen rules, 4 independent blocks, OOS blocks upstream of any tuning window) is sound. The two blocking CI failures must be fixed first; the OOS table framing and ADF table issues are the substantive changes needed before this represents a clean research artifact.

kumargu:
Pushed `9bb3dbfb` addressing the self-review:

**P0 (CI):** `cargo fmt` + `cargo clippy --workspace -- -D warnings` pass locally now. Re-ran the test suite — 31/31 pass.

**P1 (correctness):**
- ADF table swap: `EG_CRITICAL_VALUES` (Engle-Granger 2-step, k=1) → `ADF_CONSTANT_CRITICAL_VALUES` (standard ADF with constant, MacKinnon 1996). Old table was conservative (over-rejects) so the dom060 OOS lift numbers stand; only stored `adf_pvalue` fields shift. Cited the source in code.
- Added 5 validator tests for the gates in their enabled state (covers admit + reject paths for both ADF and dominance, plus the score-unavailable rejection).
- PR description updated to clarify that **both** dom050 and dom060 are in-sample on test2 and test3 (the q3_q1 sweep that produced both thresholds spans those blocks). Only test0 and test1 are truly OOS.

**P2:**
- Dead-code removal: dropped `DontDoConfig` struct and `Universe.dont_do` field.
- dom060 vs stack: kept the recommendation; the open checkbox in the test plan flags this for explicit sign-off rather than presenting it as decided. Stack does have higher min-Sharpe (+1.99 vs +1.64); my pushback was diversification (stack runs 4 active baskets in test2). I haven't quantified that further — happy to defer to your call.

**Minor:**
- `audit_bars.py` now covers Test0 and reports decision-bar (19:55–20:00 UTC) coverage in addition to full-day RTH counts. Result: 96.9 % in Test0, ≥97.7 % in Test1–Test3 across all 6 mega-caps.
- `build_per_block_fits.sh` corrected to "20" (5 × 4 blocks).
- `--as-of` future date now errors immediately with a diagnostic.
- New `quarterly_sharpe.py` reproduces the 7 quarterly Sharpe numbers in the PR description from the daily P&L in `report.tsv`.

The substantive walk-forward conclusion is unchanged: dom060 lifts baseline by +0.58 / +2.59 on the two truly-OOS blocks; mean OOS lift +1.59. Tag `wf-frozen-rule-baseline-v1` still points at `1af64ce7` (pre-fix); the PR head is `9bb3dbfb`.

kumargu:
Superseded by **#324**: lean replacement.

PR #324 lands only what main needs to run dom060 — 10 files, +803 lines, no deletions:
- Gate plumbing (the `cbdca858` Rust changes + the ADF-table correction + 5 new validator tests)
- One config TOML registering dom060 as the research candidate
- Live defaults unchanged

The 184 files in this PR (replay logs, per-block fits, dashboards, walk-forward analysis tooling, 7 alternative research configs, older research memos) stay on this branch and the tag **`wf-frozen-rule-baseline-v1`** as the evidence trail. They're not reviewable line-by-line and don't belong on main.

Walk-forward evidence summary, decision rule, alternative-variant comparison, and reject-list stability analysis are reproduced in #324's description.

PR #336, opened 2026-05-19
[codex] Harden basket paper startup observability

kumargu:
## Summary

- harden basket paper/live startup by quarantining overlay sidecar state when engine state is fresh or broker-recovered
- make `--disable-leadership-overlay` resolve to a true baseline picker with no rule-v1 sidecar state
- add high-signal `bug!` anomaly counters that also show up in logs as `bug=true bug_marker="BUG" kind=...`
- log configured capital, account equity, effective execution capital, leverage, and gross cap at startup

## Why

Before paper trading, restart safety and operator observability matter more than strategy changes. The risky path was a recovered or fresh basket engine accidentally reusing stale leadership classifier / picker sidecar state, plus disabled overlay runs still taking a stateful picker path.

## Validation

- `cargo check -p openquant-runner`
- `cargo test -p openquant-runner basket_live -- --nocapture`
- `cargo test -p openquant-runner basket_overlay_picker -- --nocapture`
- isolated replay smoke: `replay --start 2026-04-01 --end 2026-04-03 --state-path data/replay/codex_hardening_smoke/state.json --report-tsv data/replay/codex_hardening_smoke/report.tsv`
- isolated baseline replay smoke with `--disable-leadership-overlay`, confirming picker initialized as `baseline`

### DIFF
--- a/engine/crates/runner/src/basket_live.rs
+++ b/engine/crates/runner/src/basket_live.rs
@@ -1113,11 +1113,11 @@ pub async fn run_basket_live(
             ),
             Ok(false) => {
                 if can_load_sidecar_state && leadership_state_path.exists() {
-                    bug!(
-                        "leadership_sidecar_state_rebuilt",
-                        state_path = %leadership_state_path.display(),
-                        "persisted leadership sidecar state could not be reused and will be rebuilt"
-                    );
+                    move_sidecar_state_aside_if_present(
+                        &leadership_state_path,
+                        startup_state_source,
+                        "leadership_sidecar_state_mismatch",
+                    )?;
                 }
                 let warm_anchor = if last_processed_trading_day == Some(today) {
                     Some(today)
@@ -1180,12 +1180,11 @@ pub async fn run_basket_live(
         ),
         Ok(false) => {
             if can_load_sidecar_state && picker_state_path.exists() {
-                bug!(
-                    "overlay_picker_sidecar_state_rebuilt",
-                    state_path = %picker_state_path.display(),
-                    picker_id = overlay_picker.id(),
-                    "persisted overlay picker sidecar state could not be reused and will be rebuilt"
-                );
+                move_sidecar_state_aside_if_present(
+                    &picker_state_path,
+                    startup_state_source,
+                    "overlay_picker_sidecar_state_mismatch",
+                )?;
             }
             overlay_picker.save_state(&picker_state_path)?;
         }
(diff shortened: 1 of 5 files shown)


engine/crates/runner/src/main.rs
```diff
@@ -48,6 +48,14 @@ use std::io::Write;
 use std::path::PathBuf;
 use tracing::{error, info, warn};
 
+macro_rules! bug {
+    ($kind:literal, $($field:tt)*) => {{
+        metrics::counter!("bug", "component" => "basket_runtime_config", "kind" => $kind)
```
kumargu, line 53: The new `bug!` counters look useful, but right now the runner never installs a metrics recorder. In practice that means every `metrics::counter!("bug", ...)` call is a no-op outside of tests/bench code, so we only get the log line and not an actual metric stream. If the intent is to alert or count anomalies operationally, we should either install `openquant_metrics` in the runner startup path or drop the metric increment here and rely on the `BUG` log marker alone.

engine/crates/runner/src/basket_live.rs
```diff
@@ -1135,13 +1161,32 @@ pub async fn run_basket_live(
...
+        overlay_picker.load_state(&picker_state_path)
+    } else {
+        Ok(false)
+    } {
         Ok(true) => info!(
             state_path = %picker_state_path.display(),
             picker_id = overlay_picker.id(),
             "loaded persisted basket overlay picker state"
         ),
         Ok(false) => {
```
kumargu, line 1181: This mismatch/rebuild path will overwrite an existing picker sidecar in place. That is a problem for one-off baseline runs against an existing paper snapshot: if the snapshot was last used with `rule_v1`, then `--disable-leadership-overlay` resolves to a baseline picker, `load_state(...)` returns `false` because the picker ids differ, and `save_state(...)` clobbers the old rule-v1 dwell/off-signal state. I think we should quarantine mismatched picker sidecars here the same way we do for fresh/broker-recovered startup instead of rewriting them in place.

kumargu:
Review summary: I found two actionable issues in the hardening pass. First, the new `bug!` counters currently increment no-op metrics because the runner never installs a metrics recorder, so only the `BUG` log line is real. Second, the picker mismatch path can overwrite an existing `rule_v1` picker sidecar during a one-off `--disable-leadership-overlay` run instead of quarantining it, which can destroy dwell/off-signal state for the next normal paper start. I left both as inline comments with details.

PR #344, opened 2026-05-28
[codex] finalize buildout support-aware band and clean basket baselines

kumargu:
## Summary

This PR now converges on the measured production rule for basket paper/live execution:

- keep the baseline basket engine gate (`BertramFrozen`) unchanged
- add symbol-support observability so we can distinguish unsupported exits from same-side supported reallocations
- add a support-aware reallocation band to suppress only small same-side supported trims and adds
- enable that band in the buildout universe
- keep the rest of the engine honest: unsupported exits and sign flips still go through

The final scope is intentionally smaller than the research path we explored. The PR now contains the production change that survived replay, plus the baseline naming cleanup needed to make those replay comparisons intelligible.

Review follow-up: the dormant share-floor code path and log-only support-signal plumbing were removed so the PR scope matches the production behavior it describes.

## Tenets

This work was driven by four execution tenets:

1. True unsupported exits and true sign flips must still execute.
2. Small same-side reallocations should often be ignored.
3. Same-side supported symbols should not churn just because aggregation is blunt.
4. We should not stack a second opaque entry gate downstream of the engine.

A fifth pragmatic tenet emerged from the replay work:

5. Simpler is better if it preserves the intended behavior and holds up on the real replay path.

### 1. Symbol-support observability in the real runner

The basket runner now classifies the post-aggregation share book by symbol before order emission, so replay/live logs can tell us whether a proposed change is:

- `unsupported_exit`
- `supported_exit`
- `unsupported_trim`
- `supported_trim`
- `same_side_add`
- `new_entry`
- `sign_flip`

This turns the previous "aggregated" bucket into something auditable.

### 2. Support-aware reallocation band

The landed behavior is deliberately narrow:

- only applies when `supported_reallocation_band_enabled = true`
- only preserves `supported_trim` and `same_side_add`
- requires same-side structural support from admitted baskets
- respects `max_notional` and `max_shares`
- does **not** preserve unsupported exits
- does **not** block sign flips
- does **not** introduce a second entry gate

This is the smallest implementation that matches the tenets and survives replay.

### 3. Buildout universe now uses the measured band

`config/basket_universe_buildout.toml` now enables:

- `supported_reallocation_band_enabled = true`
- `supported_reallocation_band_max_notional = 1000`
- `supported_reallocation_band_max_shares = 1`

### 4. Baseline naming cleanup from earlier in the PR remains

The PR still includes the earlier naming/doc cleanup that was necessary to stop mixing `v1`, `standard`, and `buildout` language in replay discussion:

- canonical standard universe path is `config/basket_universe.toml`
- `BASELINES.md` was updated to separate `standard_core`, `buildout_core`, and `buildout_overlay`
- stale benchmark wording was replaced with reproducible buildout baseline numbers

## What We Tried And Explicitly Rejected

This is important context for review.

### Rejected: supported-flatten preservation by default

We traced the MU case deeply and proved that some sell-to-zero events were mechanically bad. But when generalized into a default production rule, preserving supported flattens degraded replay.

So the final PR does **not** preserve supported flattens by default.

### Rejected: fresh-entry confirmation / downstream entry veto

We also implemented and instrumented a fresh-entry confirmation layer. The diagnostics showed it behaved like a second downstream gate and vetoed too many engine-approved entries.

That variant regressed replay and added unnecessary complexity, so it was removed from the final path.

## Why This Shape Is Correct

The evidence from the trade audit and replay work pointed to a narrow failure mode:

- unsupported exits were generally fine
- the real churn problem was small same-side reallocations and mechanical reshaping
- broad basket-count cuts and downstream entry vetoes were too blunt

So the final implementation fixes the measured churn problem without rewriting the strategy.

### Focused tests

- `cargo test -p openquant-runner test_supported_reallocation_band_preserves_small_supported_trim_and_add -- --nocapture`
- `cargo test -p openquant-runner test_supported_reallocation_band_skips_exits_flips_and_large_changes -- --nocapture`
- `cargo test -p basket-picker test_load_universe_minimal -- --nocapture`

### Replay validation on the real basket replay path

Current-code comparison, same binary, same buildout universe, same date range (`2026-01-02` to `2026-05-27`):

- baseline with band off:
  - `cum_return = 0.518832`
  - `sharpe = 3.927919`
  - `max_dd = 0.081475`
- final buildout config with band on:
  - `cum_return = 0.536659`
  - `sharpe = 4.057172`
  - `max_dd = 0.076676`

So the landed rule improved return and Sharpe while reducing max drawdown.

## Practical Outcome

This PR does **not** claim to solve all paper-vs-replay drift.

What it does claim, with evidence, is narrower and more defensible:

- the buildout execution layer is now less willing to churn tiny same-side supported reallocations
- the runner logs are much better at telling us *why* a symbol changed
- the final rule is simpler than the broader research path
- the final rule is replay-positive on the real buildout path

That gets the portfolio-arbitration work to a clean finish line. The remaining gap is more about paper/live execution drift than unfinished strategy wiring inside this PR.


kumargu:
1. ShareFloorConfig / preserve_near_unit_shares looks like out-of-scope code. The PR description explicitly says you rejected supported-flatten preservation, and the buildout config does
  not set preserve_near_unit_shares = true. But ~80 lines of code (ShareFloorConfig, ShareFloorPreservation, TargetShareConversion, the rounding branch in target_shares_from_notionals, plus
   two tests and a logging block in process_session_close) are still in this PR, disabled. Either remove it or note clearly why it's staying behind a flag.
  
  2. build_symbol_support_snapshots is called twice in process_session_close (engine/crates/runner/src/basket_live.rs, once pre-band, once for the post-band observability log). It's
  O(symbols × selected_baskets) — fine at this scale, but the second call's class counts will look different from the first (preserved symbols flip from SupportedTrim/SameSideAdd to
  Aligned), which is easy to misread when reading logs. Worth a comment on the second build site explaining that the trade-class summary intentionally reflects the post-band plan.
  
  3. Missing doc comments on new public types. SupportedReallocationBandConfig, ShareFloorConfig, and SymbolSupportSnapshot have no /// docs. The semantics of max_notional (per-symbol
  delta, not portfolio), max_shares (per-symbol delta share count), and "same-side support" (counted from admitted baskets only, peers counted on the opposite side of target) are
  non-obvious from field names alone.
  
  4. same_side_signal_abs_sum is computed and stored on every snapshot but only ever appears in log output. If you intend to gate on signal strength later, fine — otherwise it's plumbing
  that's earning its keep only as a log field.

